PR #2, opened 2015-02-17
Add POST report endpoint

ajacksified:
:eyeglasses: @btholt


src/endpoints/v1.es6.js
```diff
@@ -312,6 +312,34 @@ class APIv1Endpoint {
     }, this)
   }
 
+  get reports () {
```
btholt: :heart: ES6

btholt:
:fish:

PR #3, opened 2015-02-20
Bump / rename 6to5 and remove dead dependency.

ajacksified:
6to5 renamed to 'babel', so this renames and upgrades.

Will require a minor version bump.

:eyeglasses: @btholt or @danehansen


btholt:
:fish:

PR #5, opened 2015-03-26
Wrap data

ajacksified:
:eyeglasses: @btholt 

Returns meta alongside data, to include extra info from the headers.


btholt:
:fish:

PR #6, opened 2015-04-01
Add subreddit API call

jo-asakura:
:eyeglasses: @ajacksified

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -6,6 +6,7 @@ import Account from '../models/account';
 import Comment from '../models/comment';
 import Link from '../models/link';
 import Vote from '../models/vote';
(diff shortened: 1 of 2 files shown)


src/endpoints/v1.es6.js
```diff
@@ -182,6 +183,27 @@ class APIv1Endpoint {
     cache.set(key, data);
   }
 
+  get subreddits () {
+    return bind({
+      buildOptions: function(options) {
+        var uri = this.origin + `/r/${options.query.subreddit}/about.json`;
```
ajacksified: :no_bell: we may need to move uri building into the actual methods (`get`, `post`) instead of in buildOptions at some point. Not a change you need to make - but this is going to get messy as we start to add more options to query subreddits.

ajacksified:
:fish:

PR #9, opened 2015-04-09
Decode APIv1's `websafe_json` encoding on JSON responses

JordanMilne:
:eyeglasses: @ajacksified 

This should fix the HTML encoding in titles and such. Wasn't able to test it with Snoode itself, but I mounted the parser on a superagent request in the REPL and it seems to work.

We'll also need to remove the `cleanHtml` calls in `-mobile` so we don't double decode HTML (and possibly XSS ourselves.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "snoode",
-  "version": "1.1.2",
+  "version": "1.2.0",
   "description": "node and browser reddit api library",
   "main": "src/api.es6.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


ajacksified:
:fish: - will do a version bump and publish, then update mobile.

JordanMilne:
Hmm, we could do that, but the problem with doing it on `.end` is we'd have to the entire object, which is kind of silly :/. Oh well!

JordanMilne:
:haircut: parse `res.text` in `.end()` instead, https://github.com/reddit/snoode/pull/9/files?w=0 to see it without whitespace changes.

I've tested it in a node REPL but IDK if it works on the client

ajacksified:
:fish:

PR #12, opened 2015-04-14
Add subreddits summary to the search; add popular subreddits end point.

jo-asakura:
:eyeglasses: @ajacksified

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -217,14 +217,19 @@ class APIv1Endpoint {
     cache.set(key, data);
   }
 
-  get subreddits () {
+  get subreddits() {
     return bind({
-      buildOptions: function(options) {
-        var uri = this.origin + `/r/${options.query.subreddit}/about.json`;
+      buildOptions: function (options) {
+        var uri = this.origin;
+        if (options.query.where) {
+          uri += `/subreddits/${options.query.where}.json`;
+        } else {
+          uri += `/r/${options.query.subreddit}/about.json`;
+        }
         return { uri, options };
       },
 
-      get: function(options = {}) {
+      get: function (options = {}) {
         var { uri, options } = this.subreddits.buildOptions(options);
 
         return baseGet(this.cache.subreddits, uri, options, this.request, (body) => {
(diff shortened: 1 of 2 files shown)


src/endpoints/v1.es6.js
```diff
@@ -217,14 +217,19 @@ class APIv1Endpoint {
...
 
-  get subreddits () {
+  get subreddits() {
     return bind({
-      buildOptions: function(options) {
-        var uri = this.origin + `/r/${options.query.subreddit}/about.json`;
+      buildOptions: function (options) {
+        var uri = this.origin;
+        if (options.query.where) {
+          uri += `/subreddits/${options.query.where}.json`;
```
ajacksified: Can we use `sort` instead of `where`, like we do for `hot` / `best` / etc elsewhere? That parameter matches more what the query is actually doing. (`gold` is weird, but such is the API.)

jo-asakura: It's reddit's API term not mine `GET /subreddits/where`

jo-asakura: `hot / best / etc` is actually `sort` param and has nothing to do with `where` :smile:

src/endpoints/v1.es6.js
```diff
@@ -234,6 +239,21 @@ class APIv1Endpoint {
             return null;
           }
         });
+      },
+
+      popular: function (options = {}) {
```
ajacksified: `popular` shouldn't be a top level endpoint - only `get`, `post`, and other HTTP verbs. This should, instead, be built in using `query` in the `get` function.

ajacksified:
:nail_care:

src/endpoints/v1.es6.js
ajacksified: https://www.reddit.com/dev/api/oauth#GET_subreddits_{where}

> GET /subreddits/where
> The where parameter chooses the order in which the subreddits are displayed.

jo-asakura: It's not `sort` it's more likely `filter` or `type`... I can do `sort`, I don't care.

jo-asakura:
:haircut: @ajacksified

ajacksified:
:fish:

PR #13, opened 2015-04-20
Add browserify babelify option to package.json

ArrestedDevelopment:
This change allows the babelify transform to be used in a browserify project (in my case, a browser extension project)

Adding `.es6.js` to `extensions` must be done on the main browserify instance (like, in the project's gulpfile). It doesn't seem to obey it if I only have it in Snoode's package.json for some reason.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,13 @@
     "url": "https://github.com/ajacksified/snoode/issues"
   },
   "homepage": "https://github.com/ajacksified/snoode",
+  "browserify": {
+    "transform": [
+      [
+        "babelify"
+      ]
+    ]
+  },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


ajacksified:
:fish: thanks!

ajacksified:
Had to revert this - it broke some builds. Will need to do more investigation to figure out what was going on here.

ArrestedDevelopment:
Really weird. Will they work if you add babelify as a dependency?

ArrestedDevelopment:
Would running babel on snoode before publishing to NPM (possibly using a [prepublish script](https://docs.npmjs.com/misc/scripts) to make things easier) break your build scripts?

PR #14, opened 2015-04-20
Add basic stylesheet support

ArrestedDevelopment:
This adds very basic support for subreddit stylesheets. I didn't add `POST` support for stylesheets in this commit, because the only kind of response one gets from `/api/subreddit_stylesheet` is meant for parsing with reddit's [`handleResponse`](https://github.com/reddit/reddit/blob/268a6297132bed80905ab8af605d93e36bd6042e/r2/r2/public/static/js/jquery.reddit.js#L126) function and it's kinda awkward to parse right now.

This allows grabbing the source of a subreddit's stylesheet (html entities unencoded) and its images.


ajacksified:
:fish: thanks!

PR #15, opened 2015-04-20
Fix double decoding of embed code

JordanMilne:
:eyeglasses: @ajacksified 

We websafe_json decode further down in the stack now.

Haven't been able to test this, but from my reading of the JSON responses and the rest of Snoode this should be correct.


ajacksified:
:fish:

PR #18, opened 2015-04-22
Pass most options (auth, origin) in through options builder

ajacksified:
:eyeglasses: @ajacksified and I guess @xilvar


ajacksified:
much simplicity wow few cache :doge: :fish:

PR #21, opened 2015-05-14
Remove unnecessary fields from json

ajacksified:
Nukes several kb of unnecessary data flushed down.

:eyeglasses: @danehansen or @curioussavage


danehansen:
:fish:

PR #28, opened 2015-06-08
Add `thingId` validator

dwick:
A `thingId` is a type prefix: `tN` and a base36 number separated by an underscore.

`t3_1` is a valid `thingId` that previously didn't meet the length requirement.

:eyeglasses: @ajacksified

### DIFF
--- a/src/models/account.es6.js
+++ b/src/models/account.es6.js
@@ -4,6 +4,16 @@ class Account extends Base {
   constructor(props) {
     props._type = 'Account';
     super(props);
+
+    var account = this;
+
+    this.validators = {
+      thingId: function() {
+        var thingId = this.get('thingId');
+
+        return Base.validators.thingId(thingId);
+      }.bind(account),
+    };
   }
 };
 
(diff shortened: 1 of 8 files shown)


ajacksified:
tests pls https://github.com/reddit/snoode/tree/master/test

dwick:
after lunch!

dwick:
:haircut: tests added

src/models/account.es6.js
```diff
@@ -4,6 +4,16 @@ class Account extends Base {
...
     props._type = 'Account';
     super(props);
+
+    var account = this;
+
+    this.validators = {
+      thingId: function() {
+        var thingId = this.get('thingId');
+
+        return Base.validators.thingId(thingId, '2');
```
ajacksified: `2` might be better off as a constant on the class (here and below models)

ajacksified: The `thingId` validator should also be optional - it should only test the regex _if_ it is not undefined, such as when creating a new Account to be `POST`ed.

ajacksified:
:nail_care:

dwick:
:haircut: didn't add the `const`s as class members because i didnt see a use case, but they're at least defined as private `const`s now.

dwick:
:haircut: removed specific type prefixes since they can be different per install.

ajacksified:
:fish:

dwick:
da68efe49f2f4c5ee45c547913ac375f3c9ff668

PR #31, opened 2015-06-17
Move ad image props into image previews

dwick:
This prevents the UI from having to handle edge cases for ad
images.

:eyeglasses: @ajacksified

### DIFF
--- a/src/models/link.es6.js
+++ b/src/models/link.es6.js
@@ -5,6 +5,34 @@ class Link extends Base {
   constructor(props) {
     delete props.selftext_html;
 
+    if (props.promoted) {
+      var resolutions = [];
+
+      if (props.mobile_ad_url) {
+        resolutions.push({
+          url: props.mobile_ad_url,
+          height: 628,
+          width: 1200,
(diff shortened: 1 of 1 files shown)


src/models/link.es6.js
```diff
@@ -5,6 +5,34 @@ class Link extends Base {
   constructor(props) {
     delete props.selftext_html;
 
+    if (props.promoted) {
```
ajacksified: An ad model would be preferable to modifying the link model (it could inherit from `Link` and call `super` in the constructor to get all of the other logic, too.)

dwick: Talked in person.  Basically the reason for not using a separate model is because this is how it works internally to reddit and if it was a different model then we'd have to handle initialization on every page `Link/Ad` are used. (comments page, listing, etc).  Doing it this removes that edge case.

ajacksified:
:fish:

dwick:
ff0eaf306b5f288f37b737bc278eb1c8b6ad8787

PR #34, opened 2015-06-24
Dead code

dwick:
:eyeglasses: @ajacksified

### DIFF
--- a/src/models/link.es6.js
+++ b/src/models/link.es6.js
@@ -84,19 +84,15 @@ class Link extends Base {
 
   get thumbnail () {
     var props = this.props;
-    var thumbnailSrc = '/img/default.gif';
 
-    if (props.thumbnail) {
-      if (props.thumbnail === 'default' || props.thumbnail === 'self' || props.thumbnail === 'nsfw') {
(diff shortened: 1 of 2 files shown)


src/models/link.es6.js
```diff
@@ -82,23 +82,6 @@ class Link extends Base {
     );
   }
 
-  get thumbnail () {
-    var props = this.props;
-    var thumbnailSrc = '/img/default.gif';
-
-    if (props.thumbnail) {
-      if (props.thumbnail === 'default' || props.thumbnail === 'self' || props.thumbnail === 'nsfw') {
```
ajacksified: Thinking it through further, it might be best to have this return `undefined` in case the `props.thumbnail` property isn't a full URL to prevent some of this weirdness sneaking through.

dwick: wouldn't that result in a 404 anyway?

ajacksified: It would be a 404 unless you explicltly returned `undefined`, which would cause it to not load a thumbnail at all instead of attempting to load a `/default` or `/self` or `/nsfw` route.

dwick:
:haircut:

ajacksified:
:fish:

PR #35, opened 2015-06-24
Add PATCH method for comments and links.

curioussavage:
:eyeglasses: @ajacksified

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -573,6 +573,38 @@ class APIv1Endpoint {
         } else {
           throw new ValidationError('Link', options.model, valid);
         }
+      },
+
+      patch: function(options={}) {
+        var uri = options.origin + '/api/editusertext';
+
+        if (!options.model) {
+          throw new NoModelError('/api/editusertext');
+        }
+        // api only supports updating selftext
+        options.model.set('selftext', options.changeSet);
+
+        var valid = options.model.validate();
+
+        if (valid) {
+          var json = options.model.toJSON();
+          options.form = {
+            api_type: 'json',
+            text: json.selftext,
+            thing_id: json.name,
+          }
+
+          return basePost(uri, options, this.request, (body) => {
+            if (body.json.errors.length === 0) {
+              var updatedLink = body.json.data.things[0].data;
+              return new Link(updatedLink).toJSON();
+            } else {
+              throw body.json.errors;
+            }
+          })
+        } else {
+          throw new ValidationError('Link', options.model, valid);
+        }
       }
     }, this);
   }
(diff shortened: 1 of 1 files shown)


src/endpoints/v1.es6.js
```diff
@@ -658,6 +690,38 @@ class APIv1Endpoint {
           defer.reject('Comment', options.model, valid);
           return defer.promise;
         }
+      },
+
+      patch: function(options={}) {
+        var uri = options.origin + '/api/editusertext';
+
+        if (!options.model) {
+          throw new NoModelError('/api/editusertext');
```
ajacksified: Should throw `api/comment` (comment being the missing model)

ajacksified:
you added a `patch` for comments, but not for selftext; those should be seperate APIs. (If you're copypasting code, create a function for the two to share.)

:nail_care:

curioussavage:
unfortunately the both use this endpoint: https://www.reddit.com/dev/api#POST_api_editusertext

src/endpoints/v1.es6.js
curioussavage: in the constructor for noModelError the variable seems to imply the endpoint is what should be supplied

curioussavage:
:nail_care:

curioussavage:
oops :haircut:

src/endpoints/v1.es6.js
curioussavage: I do think it might be nice if it actually specified the model that was not supplied

src/endpoints/v1.es6.js
```diff
@@ -903,6 +911,43 @@ class APIv1Endpoint {
...
   }
 
+  updateCommentOrLink (options) {
+    var uri = options.origin + '/api/editusertext';
+
+    if (!options.model) {
+      throw new NoModelError('/api/editusertext');
+    }
+    // api only supports updating selftext
+    var prop = options.model.get('selftext') ? 'selftext' : 'body';
```
ajacksified: It would be safer and more clear to check if `options.model._type === 'Link'` or `'Comment'` instead of checking `selftext`, which could be an empty string.

src/endpoints/v1.es6.js
```diff
@@ -903,6 +911,43 @@ class APIv1Endpoint {
...
+             return new Link(updatedThing).toJSON();
+           } else {
+             return new Comment(updatedThing).toJSON();
+           }
+        } else {
+          throw body.json.errors;
+        }
+      })
+    } else {
+      throw new ValidationError('Link', options.model, valid);
```
ajacksified: The error thrown should be `options.model._type` instead of `'Link'`

ajacksified:
> they both use this endpoint

I know :(

:nail_care:

curioussavage:
:haircut:

ajacksified:
:fish:

curioussavage:
a09ffcbb43384b98e09744b3cab31099e6a47b09

PR #37, opened 2015-06-30
edit: fix check for model type

curioussavage:
:eyeglasses: @ajacksified 

looks like I missed this somehow.


ajacksified:
:fish:

PR #38, opened 2015-07-01
add del endpoint to comment and link

curioussavage:
:eyeglasses: @ajacksified


ajacksified:
:fish: - make sure to test this in old browsers, some don't like using `delete` as a property name, since it's a keyword. (we might end up having to change it to `del` or something)

curioussavage:
So far looks good on all the browsers. Looks like babel only references it using bracket notation, shouldn't that always work?

Anyways. removed the callbacks too per our conversation yesterday :haircut:

ajacksified:
Squash and :fish:

PR #40, opened 2015-07-21
Fix response handling and request for POST message

curioussavage:
:eyeglasses: @danehansen @dwick

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -966,8 +966,8 @@ class APIv1Endpoint {
             options.form.from_sr = json.fromSr;
           }
 
-          if (json.catpcha) {
-            options.form.captcha = json.catpcha;
+          if (json.captcha) {
+            options.form.captcha = json.captcha;
           }
 
           if (json.iden) {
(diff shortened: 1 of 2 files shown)


src/endpoints/v1.es6.js
```diff
@@ -983,11 +983,17 @@ class APIv1Endpoint {
           }
 
           return basePost(uri, options, this.request, (body) => {
-            if (body) {
+            if (body.json.data) {
```
dwick: this scares me a little, sure there are no edge cases?

src/endpoints/v1.es6.js
```diff
@@ -983,11 +983,17 @@ class APIv1Endpoint {
...
               return new Message(message).toJSON();
+            } else if (body.json.captcha) {
+              return body.json;
+            } else if (body.json.errors.length){
+              throw body.json.errors;
             } else {
-              return null;
+              // for successful new messages in compose
+              // no data is returned
+              return 'success';
```
dwick: returning `null` seems better than a string that is never used.

dwick:
:nail_care:

curioussavage:
:haircut: 

added validators too

src/models/message.es6.js
```diff
@@ -4,7 +4,23 @@ class Message extends Base {
...
+        return Base.validators.minLength(this.get('subject'), 1) &&
+               Base.validators.maxLength(this.get('subject'), 100);
+      }.bind(message),
+      to: function() {
+        return Base.validators.minLength(this.get('subject'), 1);
+      }.bind(message),
+    };
   }
-};
+}
```
dwick: nit: `;`

dwick:
nit, but :fish:.  we should also figure out how to have proper status codes at some point instead of 200ing all the things

PR #43, opened 2015-08-25
Add options to load one or many links by id

ajacksified:
`options.query.id=fullname` or `options.query.ids=[fullname, ...]` for links.get

:eyeglasses: @curioussavage or @dwick


curioussavage:
lgtm :fish:

PR #51, opened 2015-10-19
Disable server cache

ajacksified:
:eyeglasses: @dwick or @curioussavage 

Also a notifaction for @spladug


dwick:
:fish:

spladug:
sorry to butt in, but i found this code really hard to understand.  the function name, `cacheCheck` doesn't tell me what the meaning of its return value is and so the boolean flippage was really confusing to follow. maybe a name like `isCacheable` would help?

ajacksified:
Sure, renaming makes sense - I'll follow up with a PR shortly.

PR #57, opened 2015-11-03
Do not delete the `subreddit` prop from comments

ajacksified:
I believe it was originally removed to save data dumped in the json of a comments page, but it's necessary in many cases (such as user activity), and can be removed in the view layer instead.

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #59, opened 2015-11-04
Add API logging capabilities

ajacksified:
Add request log if your `debuglevel` is `info`.

:eyeglasses: @curioussavage 

(Sorry for the mess, also cleaned up a little whitespace.)

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -47,19 +47,30 @@ function processMeta(headers) {
 
 const TIMEOUT = 5000;
 
-function returnGETPromise (options, formatBody) {
- return new Promise(function(resolve, reject) {
+function returnGETPromise (options, formatBody, log) {
+  return new Promise(function(resolve, reject) {
+    let time = Date.now();
+
+    log('requesting', 'GET', options.uri, options);
+
     let sa = superagent
-      .get(options.uri)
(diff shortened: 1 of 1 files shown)


src/endpoints/v1.es6.js
```diff
@@ -144,9 +155,20 @@ class APIv1Endpoint {
     });
   }
 
+  log (state, method, uri, options, status, err, duration) {
+    if (this.config.debugLevel === 'info') {
```
umbrae: I guess we don't have integer equivalents for `>=` comparisons of debug level? Do we have a "debug" level for `debugLevel`? If so, shouldn't this also still run?

ajacksified: Yeah... I suppose I should use an enum instead.

curioussavage:
is there a pr for restcache too?

ajacksified:
Correct. It may also make sense to log cache hits/misses as well, so maybe I'll follow up with one in restcache eventually too.

src/endpoints/v1.es6.js
ajacksified: Right now, it's all string based - I'll make a separate PR whereupon I overhaul the logging system.

curioussavage:
:fish:

PR #60, opened 2015-11-09
Don't follow API redirects; return a 404 instead

ajacksified:
:eyeglasses: @curioussavage 

Sometimes the API decides to send back a 302 instead of a 404; for example, unfound subs will redirect you to the search page. This kills the client.


curioussavage:
:fish:

PR #62, opened 2015-11-18
Configurable timeout

ajacksified:
Send `timeout: Xms` in the constructor's configuration object to
configure the API timeout (`timeout: 5000` is default.)

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #63, opened 2015-12-07
handle errors for comment post endpoint

curioussavage:
:eyeglasses: @ajacksified 

used lodash because its in the main repo already and should dedupe right?

also any thoughts on refactor of the this file? it would be awesome to break it out into small modules


schwers:
:fish: LGTM but I'd double check with @ajacksified on the lodash implications

ajacksified:
:fish:

PR #66, opened 2015-12-17
Update user subscription cache, in place if possible

schwers:
:eyeglasses: @ajacksified

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -403,15 +403,59 @@ class APIv1Endpoint {
           };
 
           return this.basePost(uri, options, (body) => {
+            // Update the subscriber request cache to include or exclude
+            // the new subreddit if the request didn't fail
+            if (!Object.keys(body).length) {
+              this.subscriptions.updateSubscribedCache(options, json.action);
+            }
+
             return body;
           }, 'subreddits', json.sr, {
-            user_is_subscribed: true
+            user_is_subscriber: json.action === 'sub',
           });
         } else {
           return new Promise(function(resolve, reject) {
             reject('Subscription', options.model, valid);
           });
         }
+      },
+      updateSubscribedCache: function(options, action) {
+        if (!options.id || !options.id.length || !action.length) {
+          return;
+        }
+
+        const subscribedCacheKey = `${options.origin}/subreddits/mine/subscriber.json`;
+
+        const subscribedRequestCache = this.cache.requestCache.get(subscribedCacheKey);
+
+        if (!subscribedRequestCache) { return };
+
+        // if there's not the default entry we'll assume something fancy
+        // is going on and not update the cache
+        const keys = subscribedRequestCache.keys();
+        if (keys.length !== 1) {
+          // if there's more keys just blow away the cache
+          subscribedRequestCache.reset();
+          return;
+        };
+
+        const key = keys[0];
+        const requestCache = subscribedRequestCache.get(key);
+        if (!requestCache) { return; }
+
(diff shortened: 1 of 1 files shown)


src/endpoints/v1.es6.js
```diff
@@ -399,15 +399,59 @@ class APIv1Endpoint {
...
+
+        if (!subscribedRequestCache) { return };
+
+        // if there's not the default entry we'll assume something fancy
+        // is going on and not update the cache
+        const keys = subscribedRequestCache.keys();
+        if (keys.length !== 1) {
+          // if there's more keys just blow away the cache
+          subscribedRequestCache.reset();
+          return
```
ajacksified: `;`

src/endpoints/v1.es6.js
```diff
@@ -399,15 +399,59 @@ class APIv1Endpoint {
...
+        };
+
+        const key = keys[0];
+        const requestCache = subscribedRequestCache.get(key);
+        if (!requestCache) { return };
```
ajacksified: `;`

src/endpoints/v1.es6.js
```diff
@@ -399,15 +399,59 @@ class APIv1Endpoint {
...
+
+        let subreddits = requestCache.subreddits || [];
+        const updateId = options.id;
+        if (action === 'unsub') {
+          subreddits = subreddits.filter(function(id) {
+            return id !== updateId;
+          });
+        } else {
+          if (!subreddits.some(function(id) { return id === updateId })) {
```
ajacksified: `;`

src/endpoints/v1.es6.js
```diff
@@ -399,15 +399,59 @@ class APIv1Endpoint {
...
+            // Update the subscriber request cache to include or exclude
+            // the new subreddit if the request didn't fail
+            if (!Object.keys(body).length) {
+              this.subscriptions.updateSubscribedCache(options, json.action);
+            }
+
             return body;
           }, 'subreddits', json.sr, {
-            user_is_subscribed: true
+            user_is_subscriber: json.action === 'unsub' ? false : true,
```
ajacksified: could this just be `user_is_subscriber: json.action === 'sub'`?

src/endpoints/v1.es6.js
```diff
@@ -399,15 +399,59 @@ class APIv1Endpoint {
...
             reject('Subscription', options.model, valid);
           });
         }
+      },
+      updateSubscribedCache: function(options, action) {
+        if (!options.id || !options.id.length || !action.length) {
+          return;
+        }
+
+        const subscribedCacheKey = `${options.origin}/subreddits/mine/subscriber.json`
```
ajacksified: `;`

ajacksified:
A couple of nits, but otherwise looks good. :nail_care:

schwers:
:haircut:

schwers:
@ajacksified Something I was thinking of: what are your thoughts on clearing the frontpage cache when subscriptions happen?

ajacksified:
@schwers makes sense, as the same query will return different information.

A way to abstract the logic, so the cache works automatically, would be to query by passing in an array of subreddit names (rather than *no* subreddit name), to get the same results as the front page; but I suspect the logic is different when you request the front page rather than `suba+subb+subc+subd...`.

ajacksified:
:fish:, unless you want to add in clearing out the frontpage cache

ajacksified:
re-:fish:ed

ajacksified:
https://github.com/reddit/snoode/commit/7711530638c328a10f1da6aa11490ad588fd4426

PR #68, opened 2015-12-18
Nuke event emitter and pass the linter

ajacksified:
:eyeglasses: @curioussavage or @schwers 

This massively improves performance (~25x or so) when parsing comment thread responses.


ajacksified:
:warning: currently has an issue with captchas until this message disappears

src/endpoints/v1.es6.js
```diff
@@ -111,8 +111,8 @@ function returnGETPromise (options, formatBody, log) {
 }
 
 function bind(obj, context) {
```
schwers: Can we rename this to something like `bindAll` to avoid confusion?

src/endpoints/v1.es6.js
```diff
@@ -631,14 +635,14 @@ class APIv1Endpoint {
...
-      }
+      },
     }, this);
   }
 
   get preferences () {
     return bind({
       buildOptions: function(options) {
-        var uri = options.origin + '/api/v1/me/prefs';
+        let uri = options.origin + '/api/v1/me/prefs';
```
schwers: Can this be a `const`?

src/models/base.es6.js
```diff
@@ -1,59 +1,41 @@
...
-    this.emitter = opts.emitter || new EventEmitter();
 
-    for (var p in props) {
+    for (let p in props) {
       this.props[p] = props[p];
     }
   }
 
+  get _type () {
+    return this.constructor.name;
```
schwers: This seems like it wouldn't work across minifying, do we (/ other js bundlers in general) have something a prop for this during minifying?

src/models/base.es6.js
```diff
@@ -1,59 +1,41 @@
...
-    this.emitter.emit.apply(this, args);
-  }
-
-  on (...args) {
-    this.emitter.on.apply(this, args);
-  }
-
   validate () {
-    if (!this.validators) {
+    const validators = this.validators();
```
schwers: How would you feel about a base implementation of `validators`? it looks like it's undefined in this instance.

src/models/base.es6.js
```diff
@@ -62,62 +44,50 @@ class Base {
...
-  },
+  static validators = {
+    integer: function(i) {
+      return i === parseInt(i);
+    },
 
-  min: function (i, min) {
-    return i >= min;
-  },
+    string: function(s) {
```
schwers: We could use `string(s) {...}`. I think the linter has been mad at me for not doing so in some cases but I'm not sure where it requires it

src/models/base.es6.js
```diff
@@ -1,59 +1,41 @@
-import { EventEmitter } from 'events';
+const THING_ID_REGEX = new RegExp('t\\d_[0-9a-z]+', 'i');
 
 class Base {
-  constructor (props={}, opts={}) {
+  constructor (props={}) {
     this.props = {};
```
schwers: :no_bell: I'll try to find the Google I/O talk about this, but I think we should consider using a real class here because: https://developers.google.com/v8/design?hl=en#fast-property-access. I'm happy to make up some performance tests and open a pr for it

src/models/base.es6.js
```diff
@@ -62,62 +44,50 @@ class Base {
...
-  },
+    maxLength: function (s, l) {
+      return Base.validators.string(s) && Base.validators.max(s.length, l);
+    },
 
-  regex: function(s, expr) {
-    return expr.test(s);
-  },
+    minLength: function (s, l) {
+      return Base.validators.string(s) && Base.validators.min(s.length, l);
```
schwers: For this and `maxLength`, how would you feel about changing the string check to one for `length` being defined. Then we could use it for arrays as well

schwers:
Had some questions. Some of them are longer term and could disappear as we iterate on Snoode.

I'll take another pass when captchas are working as well

:nail_care:

src/models/base.es6.js
ajacksified: It seems to get mad any time you include a function inside an object literal, but it isn't valid to do so. I think we may need to remove and fix that rule.

src/models/base.es6.js
ajacksified: Sure - probably a big enough change to be a separate PR. That would involve defining all properties for the data we expect to receive back from the API for each model, instead of cheating like this.

src/models/base.es6.js
ajacksified: It does work across minifying.

src/models/base.es6.js
ajacksified: (done)

ajacksified:
Turns out I didn't break captcha boxes with this PR: (https://github.com/reddit/reddit-mobile/blob/3839c16ecaa55045a664f82b4f831317918c0604/src/views/components/CaptchaBox.jsx#L85) 

Fixed here: https://github.com/reddit/reddit-mobile/pull/426

:haircut:

schwers:
@ajacksified wait, 25X or 25%?

ajacksified:
@schwers 25x.

src/endpoints/v1.es6.js
```diff
@@ -115,9 +119,9 @@ function returnGETPromise (options, formatBody, log) {
   });
 }
 
-function bind(obj, context) {
-  for (var p in obj) {
-    if(obj.hasOwnProperty(p) && typeof obj[p] == 'function' ) {
+function bindAll(obj, context) {
+  for (let p in obj) {
```
schwers: `let` can be `const`

schwers: :no_bell: (sort of) @curioussavage (because you're working on styleguide)

Styleguide recommends using `.forEach` (and other higher order functions) for iterating. I don't know if we do that consistently. Should we update that rule to note when it's okay to not do so?

src/models/base.es6.js
schwers: :no_bell: That's odd, I wonder if its our babel config/version. I use [babel's online repl](https://www.babeljs.io/repl) to experiment a lot and the following compiles just fine:
```
class ApiModel {
  static validators = {
    string(s) {
      return s === s.toString()
    }, 
    
    minLength: function(i, min) {
      return i >= min;
    },
    
    maxLength(i, max) {
      return i <= max;
    },
  }
}
```

src/models/report.es6.js
```diff
@@ -3,47 +3,50 @@ import Base from './base';
...
     };
   }
-};
+
+  reasonValidator () {
+    if (this.get('other_reason') && !(this.get('reason') === 'other')) {
+      return false;
+    }
+
+    const reasonValid = Base.validators.minLength(this.get('reason'), 1) &&
```
schwers: :no_bell: A validator for `lengthInRange` would be nice to add in the future if we have more of these range checks

schwers:
:nail_care: Looks great overall. Small nit but nothing major. It looks like the travis error is from failing tests, I'll be happy to f.ish after that.

ajacksified:
:haircut: - will need another pass after the big reddit-mobile linting branch is in, too.

schwers:
:fish: LGTM, let's do another pass of for the new new linting in a separate patch

curioussavage:
hooray for performance increase!

PR #70, opened 2016-01-07
Add "mobile_settings" flag to subreddit get

ajacksified:
This will cause the API response to include key color, banner, and thumbnail images for subreddit get.

:eyeglasses: @schwers


schwers:
:fish:

PR #71, opened 2016-01-11
Bug wiki discussions

curioussavage:
:eyeglasses: @ajacksified 

related to https://github.com/reddit/reddit-mobile/pull/439

* [ ] Fixes toJSON method for coments
* [ ] Uses a type passed in from app to handle links and revisions since the response looks exactly the same and can be empty

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -1273,7 +1273,7 @@ class APIv1Endpoint {
   get wiki () {
     return bind({
       buildOptions: function(options) {
-        var uri = options.origin;
+        let uri = options.origin;
         if (options.subreddit) {
           uri += `/r/${options.subreddit}/wiki/${options.path}.json`;
         } else {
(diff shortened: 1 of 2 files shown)


src/endpoints/v1.es6.js
```diff
@@ -1273,35 +1273,36 @@ class APIv1Endpoint {
...
-        var uri = options.origin;
+        let uri = options.origin;
         if (options.subreddit) {
           uri += `/r/${options.subreddit}/wiki/${options.path}.json`;
         } else {
           uri += `/wiki/${options.path}.json`;
         }
 
-        return { uri, options };
+        return uri;
```
ajacksified: All other `buildOptions` returns `uri, options`; can this be consistent?

ajacksified:
:nail_care:

src/endpoints/v1.es6.js
curioussavage: I can change it back then.

curioussavage:
:haircut:

src/endpoints/v1.es6.js
```diff
@@ -1273,35 +1273,36 @@ class APIv1Endpoint {
...
-        var uri = options.origin;
+        let uri = options.origin;
         if (options.subreddit) {
           uri += `/r/${options.subreddit}/wiki/${options.path}.json`;
         } else {
           uri += `/wiki/${options.path}.json`;
         }
 
-        return { uri, options };
+        return { uri, reqOptions };
```
ajacksified: `reqOptions` is not defined. Also, since everything returns `{ uri, options }` it may be better to change the name of the function argument instead, so you can return `{ uri, options }` like everything else.

ajacksified:
:nail_care:

src/endpoints/v1.es6.js
curioussavage: gah. thanks for noticing that. changing the arg sounds good. It wont compile when const is used instead of var as in the other places

curioussavage:
:haircut:

ajacksified:
:fish:

ajacksified:
https://github.com/reddit/snoode/commit/426148a13a0bb23668b07c32ae02e86b8e3a82e3

PR #73, opened 2016-01-21
Add a `deleted` property to comments

nramadas:
The comparison to determine whether a `Comment` is deleted isn't great, but at least the app won't have to do it.


nramadas:
:eyeglasses: @ajacksified

ajacksified:
:fish: :(

ajacksified:
:haircut: I guess we should add tests

nramadas:
Gonna close this guy since we all agree this is not always going to work.

PR #75, opened 2016-01-29
Fix hydrate to generate the same props as normal api calls

schwers:
:eyeglasses: @ajacksified


schwers:
My bad for eyeglassing, didn't realize there was CI for snoode. 

In general this PR is nice because it reduces the number of api requests we make.
Depends on [this restcache patch](https://github.com/reddit/restcache/pull/4)

ajacksified:
:fish:

PR #79, opened 2016-02-18
Add app=mweb-{env} to API calls

ajacksified:
Add an app name querystring so that it's easy to parse out when reading logs on the API side. Preferably this will be automated by a layer intercepting inbound requests rather than adding it to all clients, but this'll do for now.

:eyeglasses: @schwers @spladug

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "reddit-text-js": "0.5.4",
     "restcache": "1.1.0",
     "superagent": "1.3.0",
-    "superagent-retry": "git://github.com/ajacksified/superagent-retry#7530ab7ab24ead56de2eeecc8d31d9a86896bc5a"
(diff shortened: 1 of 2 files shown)


nramadas:
This is super nitpicky, but it might making parsing the logs even easier if we made this into two query params: `app=mweb&evn={env}`. That way filtering out mweb is one statement (`app != mweb`) instead of two, but you still maintain the ability to divide the data up further.

ajacksified:
This is per @spladug's request.

schwers:
:fish: from me on the code / testing side

spladug:
sweet, thanks! :fish: 

re: number of query params, either way is fine i suppose, but this way works with just a single simple grep: `grep app=mweb` and `grep app=mweb-client` for both narrow and combined filtering, whereas separate query params would be a little more complex but possibly easier for smarter tooling.

PR #80, opened 2016-02-18
Amend `app` querystring after the cache hash is calculated

ajacksified:
Otherwise it invalidates the client-side cache beacuse the querystring
is different between server and client.

:eyeglasses: @schwers

### DIFF
--- a/src/endpoints/v1.es6.js
+++ b/src/endpoints/v1.es6.js
@@ -68,68 +68,6 @@ function processMeta(headers, body) {
 
 const TIMEOUT = 5000;
 
-function returnGETPromise (options, formatBody, log) {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -29,7 +29,6 @@
     "superagent-retry": "git://github.com/ajacksified/superagent-retry#d39d7adbcd021d8ed09df440dba970c91fed9cd4"
   },
   "devDependencies": {
-    "babel": "^5.0.0",
```
schwers: :no_bell: updating to babel 6 as we did on reddit-mobile later for those playing along at home

schwers:
:fish:

PR #81, opened 2016-02-24
Errors inherit from a fake error class.

ajacksified:
Inherit from a 'fake error' class - babel 6 removed a shim that allowed extending native classes. This fakes the functionality of Error so that error classifications (x instanceof y) work properly.

:eyeglasses: @nramadas

### DIFF
--- a/src/errors/error.es6.js
+++ b/src/errors/error.es6.js
@@ -1,0 +1,15 @@
+export default class ErrorClass {
+  constructor (message) {
+    if (Error.hasOwnProperty('captureStackTrace')) {
+      Error.captureStackTrace(this, this.constructor);
+    } else {
+      Object.defineProperty(this, 'stack', {
+        value: (new Error()).stack,
+      });
+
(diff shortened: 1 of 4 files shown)


src/errors/error.es6.js
```diff
@@ -0,0 +1,15 @@
+export default class ErrorClass {
```
nramadas: would there be any value to using `util.inherits` to inherit the prototype from `Error`?

nramadas:
Question about inheriting, then :fish:

src/errors/error.es6.js
ajacksified: Making FakeError act as close as possible to the native Error implementation would be nice; but that's mostly what I've built anyways, grabbing stack traces and messages. Only reason not to is that browserify would have to pull in `utils`. I'll ship as-is, and if we find that something's lacking we can revisit.

nramadas: cool, i'm onboard with that.

PR #84, opened 2016-03-11
Message: parse message replies

curioussavage:
:eyeglasses: @ajacksified


curioussavage:
:haircut:

curioussavage:
How about I set the replies field to an empty array in the model when it is a string? It would make adding replies when you respond a little easier

ajacksified:
@curioussavage setting it as an array is fine. :nail_care: for that

curioussavage:
:haircut:  done

ajacksified:
:fish:

PR #85, opened 2016-03-11
SubredditRelationships: return promise for get/post/del

curioussavage:
:eyeglasses: @ajacksified or @schwers or @nramadas 

Looks like I forgot to return the promises here.


ajacksified:
:fish:

PR #86, opened 2016-03-15
v1: enable filtering by subreddit on messages endpoint

curioussavage:
:eyeglasses: @ajacksified

### DIFF
--- a/old-src/endpoints/v1.es6.js
+++ b/old-src/endpoints/v1.es6.js
@@ -1239,7 +1239,10 @@ class APIv1Endpoint {
   get messages () {
     return bindAll({
       buildOptions: function(options) {
-        const uri = options.origin + '/message/' + (options.view || 'inbox');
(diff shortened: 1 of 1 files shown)


old-src/endpoints/v1.es6.js
```diff
@@ -1239,7 +1239,10 @@ class APIv1Endpoint {
   get messages () {
     return bindAll({
       buildOptions: function(options) {
-        const uri = options.origin + '/message/' + (options.view || 'inbox');
+        const { filter, subredditName, origin, view } = options;
```
ajacksified: you can default `view` here:

`const { filter, subredditName, origin, view = 'inbox' } = options;`

ajacksified:
minor nit, :nail_care:

curioussavage:
:haircut:

ajacksified:
:fish:

PR #87, opened 2016-03-15
Turn i.imgur.com links into imgur.com links

schwers:
Spoke with Alex, we're supposed to turn i.imgur.com links -> imgur.com links. Seems like the best place to do this is at the API level. I'll follow up with comments if needed, but that's tricky. One concern with this is implementation is that it's unclear if we need to show the full link when someone edits their own post (and make sure it doesn't get PATCH'd into the prefixless version).

### DIFF
--- a/models/link.es6.js
+++ b/models/link.es6.js
@@ -1,6 +1,9 @@
 import Base from './base';
 import process from 'reddit-text-js';
 
+const I_IMGUR_DOMAIN = /i\.imgur\.com/;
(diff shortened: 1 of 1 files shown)


ajacksified:
I think this belongs in the view, not in the API library - this is business logic specifically related to how we display the link. Keeping it in the view also gets rid of the complication around patching.

schwers:
I guess what I really want is a view-model -like layer where we can make a copy of the post that has any business logic done ahead of time. toJson seemed like a good place but I can move it into helper functions. fwiw I found out patch only sends the text and self-text

PR #88, opened 2016-03-16
muted: add mute to subredditRelationships

curioussavage:
:eyeglasses: @ajacksified or @schwers or @nramadas

### DIFF
--- a/apis/messages.es6.js
+++ b/apis/messages.es6.js
@@ -50,6 +50,7 @@ export default class Messages extends BaseAPI {
             const constructor = CONSTRUCTORS[data.kind];
             const thing = new constructor(data.data).toJSON();
             if (constructor === CONSTRUCTORS.t4 &&
+                !Array.isArray(thing.replies) &&
                 typeof thing.replies === 'object' &&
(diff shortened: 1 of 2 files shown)


apis/messages.es6.js
```diff
@@ -50,6 +50,7 @@ export default class Messages extends BaseAPI {
             const constructor = CONSTRUCTORS[data.kind];
             const thing = new constructor(data.data).toJSON();
             if (constructor === CONSTRUCTORS.t4 &&
+                !Array.isArray(thing.replies) &&
```
schwers: :no_bell: So if there are replies, replies it's {thing: {replies: { data: { children: [] } } } } but if there are none its { thing: { replies: [] } }. That's odd types :/

schwers:
:fish:

apis/messages.es6.js
curioussavage: yeah... well the model changes replies to an empty array if there are none. the api returns it as an empty string, which is... weirder. All of the transformations to make them look the same are in the api layer now.

PR #92, opened 2016-03-25
Base: get login directly from this.config

curioussavage:
If the instances config is updated as in the loginAndSave
method then the requests can't get the correct origin without
accessing the config directly instead of its own instance var set
at instantiation

:eyeglasses: @ajacksified and @schwers 

It didn't look like origins array is being passed in so I removed. Fixes requests after using loginAndSave in mancy

### DIFF
--- a/apis/base.es6.js
+++ b/apis/base.es6.js
@@ -25,17 +25,6 @@ class BaseAPI {
     this.config = base.config;
     this.cache = base.cache;
     this.event = base.event;
-
-    if (base.config) {
-      this.origin = base.config.origin;
-
-      if (base.config.origins) {
-        let name = this.constructor.name.toLowerCase();
-
-        this.origin = base.config.origins[name] ||
-                      this.config.origin;
-      }
-    }
   }
 
   // Used to format/unformat for caching; `links` or `comments`, for example.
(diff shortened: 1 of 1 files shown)


apis/base.es6.js
```diff
@@ -25,17 +25,6 @@ class BaseAPI {
     this.config = base.config;
     this.cache = base.cache;
     this.event = base.event;
-
-    if (base.config) {
-      this.origin = base.config.origin;
-
-      if (base.config.origins) {
```
schwers: :no_bell: It doesn't look like we use origins. It looks like its just here to share configs between multiple api instances? Removing it seems simpler, just curious if it needs to be preserved. If it does `origin` could be a `getter`

schwers:
LGTM, just have the question about multiple `origins`, seems fine to remove though. :fish: assuming that's resolved

apis/base.es6.js
ajacksified: This was here originally so that you could, say, include an `ads` api with a separate origin. This probably needs to be refactored, as it won't really work as-is.

ajacksified:
:fish:

curioussavage:
I'm not sure why I didn't merge this one but it looks like it isn't an issue anymore since we just instantiate new apis after login. correct?

curioussavage:
@ajacksified ?

ajacksified:
@curioussavage shouldn't be necessary anymore, no.

PR #93, opened 2016-03-25
[WIP] Fixes for using SnoodeV2 in production

schwers:
I still need to add `api.hydrate` and look through caching. Otherwise this is a collection of fixes to start using SnoodeV2 for listings, search, and user activity


apis/saved.es6.js
```diff
@@ -54,23 +56,41 @@ class Saves extends BaseAPI {
     return super.del(postData);
   }
 
-  save (method, data={}) {
-    // Save, then update the data in the cache (since `saved` model changes also
-    // update `link` or `comment` models)
-    super.save(method, data);
+  updateCache (method, data) {
```
ajacksified: I think it would make sense for `updateCache` to be defined in `base` and be called by default when `save` happens, with this logic: https://github.com/schwers/snoode/blob/productionified/apis/base.es6.js#L224

schwers:
No longer needed with 3X

PR #94, opened 2016-03-28
WIP - Implement anonymous oauth token

ajacksified:
Uses a hack device-id, but hopefully soon we can drop that requirement.


phil303:
@schwers feel free to re-open if you want to take this over

PR #96, opened 2016-03-29
Update formatBody to take original method

ajacksified:
This fixes the case where DELETE doesn't send back a proper body and is
treated as an api request failure. (Since the API doesn't always send
back proper error codes, we were relying on JSON body format to
determine the error state.)

:eyeglasses: @schwers

### DIFF
--- a/apis/links.es6.js
+++ b/apis/links.es6.js
@@ -47,7 +47,7 @@ class Links extends BaseAPI {
 
   post (data) {
     const postData = {
-      api_type: 'data',
+      api_type: 'json',
       thing_id: data.thingId,
       title: data.title,
       kind: data.kind,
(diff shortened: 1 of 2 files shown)


ajacksified:
or :eyeglasses: @nramadas or @curioussavage

apis/links.es6.js
```diff
@@ -82,13 +82,15 @@ class Links extends BaseAPI {
...
-    } else {
+    } else if (method !== 'del') {
       if (body.json && body.json.errors.length === 0) {
         return body.json.data;
       } else {
         throw body.json;
       }
     }
+
+    return;
```
schwers: :no_bell: Looks fine to remove this, unless I'm missing something

schwers:
FWIW: might be nice to pass the original query in as well. In `Rules` I've been pulling the formatted method / query, etc out of the request object. But passing in the originals like this makes more sense. :fish:

ajacksified:
Yeah, not a bad idea. I may tackle that in a separate pr.

schwers:
I could add it in my patch as well; I made some changes to not let formatQuery clobber values for formatPath. I'll try to clean that up today or tmr morning

PR #100, opened 2016-04-14
Multis in snoodev2

ajacksified:
for @curioussavage

### DIFF
--- a/apis/base.es6.js
+++ b/apis/base.es6.js
@@ -143,7 +143,6 @@ class BaseAPI {
       s.set(this.buildHeaders());
 
       if (query.id && !Array.isArray(query.id)) {
-
         delete query.id;
       }
 
--- a/apis/multis.es6.js
+++ b/apis/multis.es6.js
@@ -1,0 +1,147 @@
+import BaseAPI from './base.es6.js';
+import Multi from '../models/multi.es6.js';
+
+const ID_REGEX = /^user\/[^\/]+\/m\/[^\/]+$/;
+
+class Multis extends BaseAPI {
+  static dataCacheConfig = undefined;
+
+  static mapSubreddits (subs) {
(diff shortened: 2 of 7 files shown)


ajacksified:
👓 @schwers @curioussavage

apis/comments.es6.js
```diff
@@ -73,7 +76,7 @@ class Comments extends BaseAPI {
         return body[1].data.children.map(Comments.mapReplies);
       } else if (body.json && body.json.data) {
         if (query.children) { // treeify 'load more comments' replies
-          return treeifyComments(body.json.data.things.map(mapReplies));
+          return treeifyComments(body.json.data.things.map(Comments.mapReplies));
```
curioussavage: good catch

apis/base.es6.js
```diff
@@ -286,6 +293,35 @@ class BaseAPI {
     return this.save('patch', data);
   }
 
+  // Get the source, then save it, modified by data.
+  copy (fromId, data) {
```
curioussavage: could you explain these in person?

apis/multis.es6.js
```diff
@@ -0,0 +1,147 @@
...
+      });
+    } else if (body) {
+      const multi = body.data;
+      multi.subreddits = Multis.mapSubreddits(multi.subreddits);
+      return new Multi(multi);
+    }
+  }
+
+  formatData (data) {
+    return Multis.formatData(data);
```
curioussavage: is using a static method for perf?

ajacksified: More because it doesn't need access to the  instance, so it doesn't need to be non-static.

apis/base.es6.js
ajacksified: Sure, anything in particular you want to cover?

PR #103, opened 2016-04-29
Reexpose html fields and avoid client-side rendering

prashtx:
We want to rely on server-rendered HTML for things like selftext, so we need to add those fields back to the model. This will let us fix a mobile-web bug with strikethrough rendering: we'll use the correctly-rendered HTML that we already have. Eliminating client-side rendering also lets us reduce dependencies.

We eliminate the `bodyHtml` getter from `Comment`, and the corresponding change to reddit-mobile will remove references to that field.

👓 @schwers @ajacksified

### DIFF
--- a/models/comment.es6.js
+++ b/models/comment.es6.js
@@ -1,21 +1,20 @@
 import Base from './base';
-import process from 'reddit-text-js';
 
 class Comment extends Base {
   _type = 'Comment';
 
   validators () {
-    const body = this.bodyValidator.bind(this);
+    const body_html = this.bodyHtmlValidator.bind(this);
     const thingId = this.thingIdValidator.bind(this);
 
     return {
-      body,
+      body_html,
       thingId,
     };
   }
 
-  bodyValidator () {
-    return Base.validators.minLength(this.get('body'), 1);
+  bodyHtmlValidator () {
+    return Base.validators.minLength(this.get('body_html'), 1);
   }
 
   thingIdValidator () {
@@ -23,14 +22,9 @@ class Comment extends Base {
     return Base.validators.thingId(thingId);
   }
 
-  get bodyHtml () {
-    return process(this.get('body'));
-  }
-
   toJSON () {
     let props = this.props;
     props._type = this._type;
-    props.bodyHtml = this.bodyHtml;
     props._type = this._type;
     return props;
   }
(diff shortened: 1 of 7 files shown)


models/comment.es6.js
```diff
@@ -1,36 +1,30 @@
 import Base from './base';
-import process from 'reddit-text-js';
 
 class Comment extends Base {
   _type = 'Comment';
 
   validators () {
-    const body = this.bodyValidator.bind(this);
+    const body_html = this.bodyHtmlValidator.bind(this);
```
ajacksified: This should be camelCased instead - we minimize use of underscores to when we're dealing directly with the raw API results.

prashtx: If I understand correctly (looking at https://github.com/reddit/node-api-client/blob/master/models/base.es6.js#L43), the validator names need to match the property names. In this case, the property name is the result of raw API data (as opposed to the previous `bodyHtml` field). Should we prefer to replace underscored API fields with camelCase fields in the model, if we're not doing any processing?

models/subreddit.es6.js
```diff
@@ -38,10 +38,6 @@ class Subreddit extends Base {
   static fields = EDIT_FIELDS;
 
   constructor(props) {
-    delete props.submit_text_html;
-    delete props.description_html;
-    delete props.public_description_html;
-
```
ajacksified: you should delete the markdown properties instead, now that we're using the html - that will reduce the size of the json payload written to the page.

ajacksified:
With this change, should we also be adding `raw_json=1` to the querystring of requests for these models, so we're not dealing with encoded html?

💅

models/comment.es6.js
ajacksified: whoops, ignore me, my head was in v2land.

prashtx:
Switched to using `raw_json=1` rather than client-side interpretation of the websafe_json.

Deleted markdown model properties in favor of html properties except for `link.selftext`. Deleting that looks like it will interfere with submission of a selftext post.

💇

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@r/api-client",
-  "version": "1.23.5",
+  "version": "1.23.6",
```
ajacksified: nit: let's make this `1.24.0`, since it's a more impactful change than just a bugfix.

prashtx:
💇

schwers:
Code looks good, for testing purposes, is this on staging?

ajacksified:
:fish:

schwers:
:fish:

PR #105, opened 2016-05-06
Turn loadmore into a property instead of a reply

ajacksified:
👓 @schwers


curioussavage:
why do we commit the bundle with webpack?

schwers:
@curioussavage it's nicer for anyone who isn't using webpack in their stack, and it makes having `npm link`'ed modules easier to develop as there's no need to resolve plugins. Still looking into the right way to build only the parts your reddit project needs

schwers:
:fish:

curioussavage:
kool

PR #107, opened 2016-05-06
Add submissionType to Subreddit model

birakattack:
Removed content_options because it is no longer returned by the API. It is not called submission_type and this property was added to the Subreddit model.

The api-client now returns the submissionType in the Subreddit model. 

:eyeglasses: @schwers


schwers:
:fish: 🐠  🍹

birakattack:
d936521

PR #110, opened 2016-05-12
expose rawSend

schwers:
👓 @ajacksified


apiBase/APIRequestUtils.es6.js
```diff
@@ -48,7 +48,13 @@ const requestHeaders = (apiOptions) => {
 };
 
 const requestPath = (apiOptions, path) => {
-  return `${apiOptions.origin}/${path}`;
+  let slash = '/';
```
ajacksified: `const slash = path.indexOf('/') ? '/' : ''` if you want to be super confusing :trollface:

ajacksified:
:fish:

PR #111, opened 2016-05-17
Support user info for logged-out users

prashtx:
We need to know the assigned experiment variants for logged-in and
logged-out users. The API will expose the data in compatible ways on
both /api/v1/me and /api/me.json, the latter for logged-out users.


prashtx:
👓 @schwers @ajacksified

schwers:
:fish:

PR #113, opened 2016-05-17
Convert Accounts endpoint in 3X

schwers:


apis/accounts.es6.js
```diff
@@ -1,27 +1,26 @@
...
 
-  path(method, query={}) {
-    if (query.user === 'me') {
-      return 'api/v1/me';
-    } else {
-      return `user/${query.user}/about.json`;
-    }
+const getPath = (query) => {
+  if (query.user === 'me') {
+    return 'api/v1/me';
```
prashtx: Do we also need to support the logged-out scenario (as with the v1 code here: https://github.com/reddit/node-api-client/blob/master/old-src/endpoints/v1.es6.js#L1039), so that we can inspect features/variants? Or should we have a separate mechanism for that in 3X?

schwers: Good point, I haven't added it in yet, but I'll pull it in

apis/BaseContentEndpoint.js
```diff
@@ -48,80 +47,49 @@ export const formatBaseContentQuery = (query, method) => {
...
-  static move = BaseEndpoint.notImplemented('move');
-  static copy = BaseEndpoint.notImplemented('copy');
-  static put = BaseEndpoint.notImplemented('put');
-
-  static formatQuery(query, method) {
-    if (method !== 'patch') {
-      query.feature = 'link_preview';
-      query.sr_detail = 'true';
-    }
+export const  patchPath = () => {
```
ajacksified: doublespace 😱

apis/BaseContentEndpoint.js
```diff
@@ -48,80 +47,49 @@ export const formatBaseContentQuery = (query, method) => {
...
+  return 'api/editusertext';
+};
 
-    if (method === 'del') {
-      query._method = 'post';
-    }
+export const  deletePath = () => {
```
ajacksified: here too 🙀

apis/BaseContentEndpoint.js
```diff
@@ -48,80 +47,49 @@ export const formatBaseContentQuery = (query, method) => {
...
-    } else if (method === 'del') {
-      return this.deletePath(query);
-    }
-  }
+  const promises = [];
 
-  static patchPath () {
-    return 'api/editusertext';
-  }
+  Object.keys(data).map(k => {
```
ajacksified: 🔕  Did you want to swap map/each/etc out for lodash for performance?

apis/accounts.es6.js
```diff
@@ -1,27 +1,32 @@
...
-      return 'api/v1/me';
-    } else {
-      return `user/${query.user}/about.json`;
-    }
+const getPath = (query) => {
+  if (query.loggedOut) {
+    return 'api/me.json';
+  } else if (query.user === 'me') {
+    return 'api/v1/me';
+  } else {
```
ajacksified: no `else` because the above statements all return. (that rule is in the linter)

ajacksified:
couple of linternits, otherwise looks good. 💅

apis/accounts.es6.js
schwers: technically `no-else-return` isn't in the `.eslintrc` in this repo. I don't have the reddit-mobile linter globally so there's some linter errors slipping by, sorry :/ After I clean up this patch I'll make a separate pr to pull in the latest linter.

apis/BaseContentEndpoint.js
schwers: I've been using the lodash methods in my own code, wouldn't be a bad idea if everyone can palette using them. Tbh I was just moving code around to get rid of the class (that removed ~ 2KB iirc, but the model took that space back)

schwers:
💇

apis/BaseContentEndpoint.js
ajacksified: 👍 , later

ajacksified:
:fish:

PR #114, opened 2016-05-20
Convert Subscriptions Endpoint and other changes for subreddit headers

schwers:


schwers:
👓 @ajacksified

apiBase/Model.es6.js
```diff
@@ -160,11 +160,14 @@ export default class Model {
...
   // };
   // ```
-  stub({ keyOrObject, value}, promise) {
-    const next = {...this.toJSON(), ...this._diff(keyOrObject, value) };
+  stub(keyOrObject, valueOrPromise, promise) {
+    if (!promise) {
+      promise = valueOrPromise;
+    }
+
+    const next = {...this.toJSON(), ...this._diff(keyOrObject, valueOrPromise) };
```
ajacksified: space between `{.`

ajacksified:
:fish: after the nit. I feel very strongly that a `get subscriptions` should be used rather than using `mine` as a sort in subreddits, but it can be done outside of this pr.

PR #115, opened 2016-05-21
Voting and commenting for posts and comments

ajacksified:
👓 @schwers


models2/CommentModel.es6.js
```diff
@@ -2,6 +2,8 @@ import RedditModel from './RedditModel';
 import Record from '../apiBase/Record';
 import { COMMENT, COMMENT_LOAD_MORE } from './thingTypes';
 
+import votable from './mixins/votable';
```
ajacksified: How do we feel about this style of adding functionality? We could also consider decorators as a cleaner syntax.

models2/mixins/votable.js
```diff
@@ -0,0 +1,31 @@
...
+
+export function downvote (apiOptions) {
+  // If already downvoted, cancel out the upvote.
+  return this._vote(apiOptions, this.likes === -1 ? 0 : -1);
+};
+
+export function _vote (apiOptions, direction) {
+  const oldModel = this;
+
+  const stub = this.stub('likes', direction, async (resolve, reject) => {
```
schwers: 🔕 💯

schwers:
:fish:

models2/CommentModel.es6.js
schwers: I think this is fine. I'm hesitant to add decorators to node-build yet as the spec is in flux ([more info](http://babeljs.io/docs/plugins/transform-decorators/)). From what I've read the new spec isn't completely backwards-compatible, and adding the babel-plugin for the 'legacy' syntax can interact with the static properties plugin in obscure, error'ing, ways. It might be better now so maybe thats worth re-considering

PR #117, opened 2016-05-24
Expose the id's for pagination to api consumers.

schwers:
Subreddit’s uuid is the ‘url name’ because of who they’re parsed /
looked up in most clients. This makes paging difficult, and there were
workarounds internally to api-client. This causes problems when
api-client consumers want to do their own subreddit pagination with
records directly. By exposing the id of records for use in pagination,
this problem is fixed.


schwers:
👓 @ajacksified

ajacksified:
> id's

apiBase/Record.es6.js
```diff
@@ -1,6 +1,7 @@
 export default class Record {
-  constructor(type, uuid) {
+  constructor(type, uuid, paginationId) {
```
ajacksified: how about defaulting:
`constructor(type, uuid, paginationId=uuid)`

apiBase/Model.es6.js
```diff
@@ -179,12 +180,16 @@ export default class Model {
     return fakeUUID();
   }
 
+  makePaginationId(data) {
```
ajacksified: why not use a getter for `paginationId` instead?

ajacksified:
just a couple of minor questions, 💇

apiBase/Record.es6.js
schwers: Sure, probably useful for anyone using Record directly.

apiBase/Model.es6.js
schwers: I want to make sure it has access to the raw api response, and that's not being stored. So making it a function was the easiest way to do it

schwers:
💇

ajacksified:
:fish:

PR #119, opened 2016-06-07
Add peerDependencies as devDependencies for convenience

schwers:
This is a bit of a workaround, but should simplify local development as peerDependencies don't need to be installed manually.

👓 @uzi


uzi:
🐠

PR #125, opened 2016-06-21
Refactor APIRequestUtils straw man

phil303:
This is about 75 lines shorter and I think utilizes promises in a more correct manner. We're now minimizing the amount of callback passing and the number of arguments since these are available in the scope of the promise's callbacks. This makes for cleaner and easier to understand code. This is just a straw man and has not been tested but I want to get your thoughts before I dive into a full refactor.

Some more notes on this:

1. I removed EventEmitter because its utility wasn't obvious to me. It can be added back in if you feel strongly.
2. I couldn't figure out where defaultErrorHandler was supposed to be defined so I removed it. Probably simple to add back as well.
3. One problem with this approach is that it conflates the idea of an http error and an exception. For now, I acknowledge that and will clean it up but I wanted to show you what I think is a more correct and cleaner approach to promises before I got into the nitty and gritty details.

:eyeglasses: @schwers

### DIFF
--- a/apiBase/APIRequestUtils.es6.js
+++ b/apiBase/APIRequestUtils.es6.js
@@ -4,14 +4,6 @@ import ValidationError from './errors/ValidationError';
 import NoModelError from './errors/NoModelError';
 import APIResponse from './APIResponse';
 
-import Events from './Events';
-
-const EventEmitterShim = {
-  emit: () => {},
-  on: () => {},
-  off: () => {},
-};
-
 const DefaultOptions = {
   userAgent: 'snoodev3',
   origin: 'https://www.reddit.com',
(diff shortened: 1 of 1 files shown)


apiBase/APIRequestUtils.es6.js
```diff
@@ -4,90 +4,54 @@ import ValidationError from './errors/ValidationError';
...
     if (s.redirects) {
       s.redirects(0);
     }
   }
 
-  s.end((err, res) => {
-    // handle super agent inconsistencies
-    const req = res ? res.request : fakeReq;
-    cb(err, res, req);
+  return new Promise((resolve, reject) => {
```
phil303: Reading through the superagent docs. This can be made even more concise, since the library supports the promise interface.

schwers:
Overall I like this approach and think its worth pursuing further. If we invest time into doing I think we should try removing the distinction of `save` and `runQuery` and consolidating them into one function (for discussion lets just call it `request`, not sold on a name though). The difference between `save` and `runQuery` seems unhelpful at the moment and the naming doesn't make it any better. With `request` we could provide `get`, `post`, `put`, `del` methods on top of `request` that could handle some if the differences and make the consuming code a little nicer to read. e.g. `get` could handle passing `kind='query'` to `request` and adding `raw_json=1` to the query.

Re more notes:

1. I think we should keep `EventEmitterShim` for now as it provides a lot of debugging value to `./repl`. Note that this doesn't use `EventEmitter` so we don't have to include the source or polyfill. I have an idea to re-write `./repl` to be more helpful that wouldn't require this at all, so if you're okay with leaving it in for now I can prototype the idea and we could potentially remove `EventEmitterShim` for good.
2. `defaultErrorHandler` was supposed to be a general hook for app's to do things like display error messages or redirect etc if there was any error. I'm not sure its needed so I'm okay with removing it for now. I'm going to be adding error pages to 2X this week so my opinion may change but I'd be happy to add it back in some form if it's needed.
3. Yup, this is fine for an example. Just note it is important that we can differentiate between these kind of errors. In the completed version errors from the API should be throwing `ResponseErrors` so we can track API errors separately from our own exceptions. For background: (Error tracking in mweb 1x)[https://github.com/reddit/reddit-mobile/blob/4926842296c4113f01413c662a437502870bd160/src/lib/errorLog.es6.js#L50] and (some error parsing in api-client v1)[https://github.com/reddit/node-api-client/blob/master/old-src/endpoints/v1.es6.js#L161]

apiBase/APIRequestUtils.es6.js
```diff
@@ -99,72 +63,30 @@ export const validateData = (data, method, apiName, validator) => {
...
 
-  getEmitter(apiOptions).emit(Events.response, req, res);
-
-  const apiResponse = tryParseResponse(reject, res, req, method, query, parseBody, parseMeta);
-
-  getEmitter(apiOptions).emit(Events.result, apiResponse);
-  resolve(apiResponse);
+  return rawSend(apiOptions, method, path, query, 'query')
+    .then((res, req) => makeApiResponse(res, req, method, query, parseBody, parseMeta))
+    .catch((err, req) => console.trace(err));
```
schwers: 🔕  Will `.catch` squash the error? In whatever form this ends up looking like, its probably important to be able to catch these errors in the calling code.

apiBase/APIRequestUtils.es6.js
schwers: I _think_ we might have had some trouble with that when we tried using promises in reddit-mobile 1x, but that would be great if it works. We ended up with https://github.com/reddit/reddit-mobile/blob/master/src/lib/makeRequest.es6.js, @nramadas probably knows how well the interface works or not

apiBase/APIRequestUtils.es6.js
phil303: So I tacked that on there for feature parity but I'm glad you caught that. Ideally, the calling code applies the catch so it can be handled more precisely there. The problem with that, as I'm sure you know, is that Promises will silently squash exceptions if you forget to add a `catch`. I'm not sure how/if we've remedied that somehow.

apiBase/APIRequestUtils.es6.js
curioussavage: yes. We looked into this before. Reading the docs it is a little confusing because it seems to indicate it does, however I found [this issue](https://github.com/visionmedia/superagent/issues/957) which  confirms that they only support promises in 2.0 ( we are on 1.3) 

I believe we found out that one of our other libraries was transforming our code to be compatible with promises ( facebook regenerator lib or something). which seems sketchy. We should update the dep or stick with the how its written now I think.

phil303: ah thanks for the context @curioussavage. I was asking Ryan but maybe you know more details, were there any efforts to use `fetch`?

apiBase/APIRequestUtils.es6.js
phil303: Talking to Ryan, it sounds like generally speaking we just `try/catch` coupled with `awaits`. There's a couple situations when this doesn't work and we'd have to lean on education/process to ensure errors are always handled, namely non-async functions that still want to use a promise and not waiting on the promise. I don't have a feel for how often that comes up in practice in our codebase.

apiBase/APIRequestUtils.es6.js
curioussavage: We have talked about it a few times I think. Can't remember all the issues. The major reason was inconsistent browser/node support I think ( could be wrong)

apiBase/APIRequestUtils.es6.js
```diff
@@ -4,96 +4,55 @@ import ValidationError from './errors/ValidationError';
...
-    const req = res ? res.request : fakeReq;
-    cb(err, res, req);
+  return new Promise((resolve, reject) => {
+    s.end((err, res) => {
+      // handle super agent inconsistencies
+      const fakeReq = { origin, path: _path, url, method, query: { ...data} };
+      const req = res ? res.request || res.req : fakeReq;
+
+      if (err) {
+        reject(err, req);
```
nramadas: i didn't think you could have multiple values resolved or rejected

phil303: actually that's a good catch, i don't think you can. Will fix it the cleaned up version.

apiBase/APIRequestUtils.es6.js
phil303: node/certain browsers don't support it yes. That can be handled with a polyfill though. Ryan mentioned that there might've been a problem with the polyfill?

phil303:
superseded by https://github.com/reddit/node-api-client/pull/141

PR #126, opened 2016-06-22
Fix comment permalink generation on vote / reply

schwers:
This fixes the failing test in mweb 2X, I've been running it locally and forgot to make the PR :( but yey tests for reminding us :)

👓 @nramadas or @phil303  or @uzi


uzi:
:fish:

PR #127, opened 2016-06-22
Fix thumbnail url parsing

schwers:
1X api client used to strip out `post.thumbnail`s that weren't urls, and were hints at how to render. This makes it easier to do our fancier preview image generation and makes the api a little more consistent. This restores that behavior in 2X


schwers:
👓 @phil303 or @uzi

models2/PostModel.es6.js
```diff
@@ -5,6 +5,14 @@ import replyable from './mixins/replyable';
 
 const T = RedditModel.Types;
 
+const cleanThumbnail = thumbnail => {
+  if (thumbnail && (thumbnail === 'default' || thumbnail === 'self' || thumbnail === 'nsfw')) {
```
uzi: Make an array with `['default', 'self', 'nsfw']` and `ARRAY.indexOf(thumbnail) !== -1`?

phil303:
```js
const cleanThumbnail = thumbnail => {
  const IGNORE = ['default', 'self', 'nsfw'];   // or use a set
  return thumbnail && IGNORE.indexOf(thumbnail) > -1 ? '' : thumbnail;
};
```

uzi: Let me 🔕  that one

schwers:
💇

phil303:
:fish: :fish:

models2/PostModel.es6.js
```diff
@@ -130,6 +136,10 @@ export default class PostModel extends RedditModel {
         }],
       };
     },
+
+    thumbnail(data) {
```
phil303: actually, shouldn't this be `thumbnail`

uzi:
:fish:

PR #130, opened 2016-06-23
Fix over18 flag in Account model

schwers:
👓  @phil303


phil303:
:fish: once CI is happy

PR #132, opened 2016-06-27
Add preferences endpoints, add the over18 flag to Posts.

schwers:
There were some minor changes I had to make in order to do the real work of this patch. `save` (slated to be renamed) needed support for patching with json bodies. Also non-GET requests should be sending the app parameter as query params for tracking purposes to have parity with 1X.

I also had to rename some files as I started using my real work laptop again -- IT reformatted it relatively recently and made it case sensitive.

👓 @nramadas @phil303


models2/Preferences.es6.js
```diff
@@ -0,0 +1,118 @@
...
+
+  static PROPERTIES = {
+    affiliateLinks: T.bool,
+    allowClicktracking: T.bool,
+    beta: T.bool,
+    clickgadget: T.bool,
+    collapseReadMessages: T.bool,
+    compress: T.bool,
+    credditAutorenew: T.bool,
+    defaultCommentSort: T.string, // It would be nice to have T.oneOf like react here
```
nramadas: we could easily add it. something to consider though, what should we see here if the value doesn't validate? an empty string, or a null value?

apis/PreferencesEndpoint.es6.js
```diff
@@ -0,0 +1,26 @@
+import { runQuery, runForm } from '../apiBase/APIRequestUtils';
+
+import Preferences from '../models2/Preferences';
+
+const prefsUrl = '/api/v1/me/prefs';
```
phil303: this should be all caps

apiBase/APIRequestUtils.es6.js
```diff
@@ -77,12 +77,12 @@ export const rawSend = (apiOptions, method, path, data, kind, cb) => {
   let s = superagent[method](url);
   s.set(requestHeaders(apiOptions));
 
-  data.app = appParameter(apiOptions);
-
-  if (kind === 'form') {
-    s.type('form');
+  if (type !== 'query') {
```
nramadas: small nitpick, but can we re-order this if? `if (type === 'query') { ... } else { ... }`

models2/Preferences.es6.js
```diff
@@ -0,0 +1,118 @@
...
+    highlightControversial: T.bool,
+    highlightNewComments: T.bool,
+    ignoreSuggestedSort: T.bool,
+    labelNsfw: T.bool,
+    lang: T.string,
+    legacySearch: T.bool,
+    markMessagesRead: T.bool,
+    media: T.string, // Another case for T.oneOf,
+    // 'on', 'off', 'subreddit'
+    minCommentScore: T.number, // T.number should maybe have a T.range version
```
nramadas: same question as with typeOf. we don't want to throw an error (i don't think. i dunno, maybe we do?), but what would this value be if it doesn't validate? NaN?

apiBase/APIRequestUtils.es6.js
```diff
@@ -105,8 +105,10 @@ export const validateData = (data, method, apiName, validator) => {
 export const runForm = (apiOptions, method, path, data, parseBody, parseMeta) => {
   if (!(apiOptions && method && path && data)) { throw new NoModelError(); }
 
+  const type = method === 'patch' ? 'json' : 'form';
```
phil303: seems like we're overloading `runForm` here

nramadas:
one small nit, otherwise looks good :nail_care:

apis/PreferencesEndpoint.es6.js
```diff
@@ -0,0 +1,26 @@
...
+
+// We opt-out of using the automatic parsing from `runQuery` and `runForm`,
+// because the preferences object doesn't really make sense in the normalized
+// response model.
+
+export default {
+  get: async (apiOptions) => {
+    const responseBody = await runQuery(apiOptions, 'get', prefsUrl, {}, {});
```
phil303: :no_bell: can we assume something called `runQuery` is always going to be a GET?

phil303:
:nail_care: some higher level questions but looks functionally good

apiBase/APIRequestUtils.es6.js
schwers: ya, sure thing

models2/Preferences.es6.js
schwers: It's a good point, I would propose having it take an optional default value, otherwise undefined. I mostly wanted to this comment in place to document future improvements we could do that I was putting off. Do you think its better to remove it?

apiBase/APIRequestUtils.es6.js
schwers: I can move it into another method. At first I didn't want to as I was hoping we'd get to the request'ing refactor sooner, but seeing it here makes me want to just move it.

apis/PreferencesEndpoint.es6.js
schwers: I think so, there are few places using for `posts` that should be switched.

apiBase/APIRequestUtils.es6.js
phil303: I'll get to it soon! (ish)

models2/Preferences.es6.js
nramadas: nah, the comment is good. it'll remind us to revisit this later

schwers:
💇

nramadas:
:fish: from me

phil303:
:fish:

PR #134, opened 2016-07-05
Fix uses of validateData now that its fixed.

schwers:
These uses of validateData were improperly written (they assumed validator should return `true` if the data was _invalid_ not valid). Fixes that so voting / subscribe works in 2X again.

👓 @nramadas


nramadas:
:fish:

PR #137, opened 2016-07-14
Stop posting endpoint from throwing invalid exceptions

phil303:
The posting endpoint wrapper was throwing exceptions when json errors were present (not 4xx's/5xx's but a 200). Since we actually need to look at the errors stop this from happening.

I handled this by not having a default `parseBody` function but allowing the mobile code to handle the promise directly.

:fish: @schwers


schwers:
💅  We talked about this in person, but just wanted to leave a note here. Things to try were testing if '.json' on the endpoint made error parsing easier, and seeing how throwing a BadCaptcha error here felt.

phil303:
@schwers :haircut:

apiBase/errors/NotImplementedError.js
```diff
@@ -1,5 +1,3 @@
-import FakeError from './FakeError';
```
schwers: After the issue with `BadCaptchaError`, should we make this extend `FakeError` as well so client's can use `instanceof`?

phil303: oh probs. I made that change before we figured that out. Lemme fix

apiBase/errors/ValidationError.js
```diff
@@ -1,20 +1,19 @@
 import FakeError from './FakeError';
 
-const msgText = (apiName, errors, model) =>
-  `${ apiName } had errors in ${ errors.join(',') } with properties ${ JSON.stringify(model.toJSON()) }`;
+const msgText = (api, errors) => `${api} had errors in ${errors.join(',')}`;
 
 export default class ValidationError extends FakeError {
-  constructor (apiName, model, errors) {
+  constructor (apiName, errors, status) {
```
schwers: 🔕 We should tweak some combination of validators and the validation in `APIRequestUtils` to give more helpful errors here.

apis/PostsEndpoint.js
```diff
@@ -56,30 +60,19 @@ const formatPostData = (data)=> {
...
+  if (data && data.children && data.children[0]) {
+    if (data.children.length === 1) {
+      apiResponse.addResult(PostModel.fromJSON(data.children[0].data));
       return;
     } else {
-      throw body.json;
+      data.children.forEach(c => apiResponse.addResult(PostModel.fromJSON(c.data)));
+      return;
     }
+  } else if (data) {
```
schwers: Looks like we can remove this now that this method only handles `GET`s

apiBase/errors/ValidationError.js
phil303: Yea, so i remove model because I can imagine this being used for things that don't have a model. The `msgText` thing can definitely be modified to be more helpful though. I do want to give devs access to the raw errors though.

apis/PostsEndpoint.js
```diff
@@ -56,30 +60,19 @@ const formatPostData = (data)=> {
...
-        data.children.forEach(c => apiResponse.addResult(PostModel.fromJSON(c.data)));
-        return;
-      }
-    } else if (data) {
-      return;
-    }
-  } else if (method !== 'del') {
-    if (body.json && body.json.errors.length === 0) {
-      apiResponse.addResult(body.json.data);
+const handleGet = (res, apiResponse, req, method) => {
```
schwers: 🔕 Tiny nit, the linter would complain about `req` and `method` being unused if it were in working order. I feel like method signatures is the one place where its okay to write out the full argument list so its easier to modify this code in the future without having to lookup what the function is being passed. How do you feel about that rule in cases like this, or any other edge cases?

apis/PostsEndpoint.js
phil303: that was in the GET block in the previous code so I didn't want to mess with it. I have no idea what it's intended for though.

schwers: My guess is it was a bad translation of https://github.com/reddit/node-api-client/blob/master/old-src/endpoints/v1.es6.js#L872, seems safe to remove

schwers:
:fish:

schwers:
:fish: 🐠

PR #138, opened 2016-07-14
No target blank automatically on links from server html

schwers:
Forgot to make a pr for this along with the UGC wrapper PR. Right now we manually force all links to be target=_blank links in html we get from the server. This breaks link hijacker, and in general probably isn't desired behavior. If we were to do this we should be using the user's preference anyway (either here, or in client code).

👓 @phil303


phil303:
:fish:

PR #139, opened 2016-07-20
Throw ResponseErrors similiarly to 1X

schwers:
👓  @phil303


apiBase/APIRequestUtils.es6.js
```diff
@@ -145,38 +149,60 @@ const normalizeRequest = (res, req) => {
...
-  const apiResponse = tryParseResponse(reject, res, req, method, query, parseBody, parseMeta);
+  const apiResponse = tryParseResponse(reject, res, req, method, path, query, parseBody, parseMeta);
 
   getEmitter(apiOptions).emit(Events.result, apiResponse);
   resolve(apiResponse);
 };
 
-const tryParseResponse = (reject, res, req, method, query, parseBody, parseMeta) => {
+const handleRequestIfFailed = (apiOptions, err, res, req, method, path, reject) => {
+  if (!(err || (res && !res.ok))) { return; }
```
phil303: these kinds of conditions are really hard to understand:
How about:
```
if (!err || (res && res.ok)) { return; }
```

apiBase/APIRequestUtils.es6.js
```diff
@@ -145,38 +149,60 @@ const normalizeRequest = (res, req) => {
...
+    }
+
+    // This is a good example of error-handling we should reconsider soon.
+    // The api sending back a 302 for a subreddit might be a good indicator
+    // that we should really be issuing a 302 to a url that has the canonical
+    // capitalization of a subreddit for SEO or any number of other things.
+    // 1X's comment:
+    // Sometimes the API decides to send back a 302 instead of a 404; for
+    // example, unfound subs will redirect you to the search page.
+    if (method === 'get' && status === 302) {
```
phil303: This makes me nervous. Is there ever a case where the 302 is real and that we should follow it to some resource? I guess that'd be captured by the error checking above?

phil303:
In general, having to pass another argument around feels like code smell to me that can be easily remedied with a promise oriented refactor. This might be asking too much, but since you're in here what do you think about doing that refactor pass we talked about first?

apiBase/APIRequestUtils.es6.js
schwers: It makes me nervous as well, I'd assume that there are times when a 302 really means 302 but the fact the error-handling was applied broadly seems important. The only thing I could find for more context was the PR for that line, https://github.com/reddit/node-api-client/pull/60. The error checking above wouldn't be handling it.

schwers: Does it seem reasonable to keep it this way for response parity with 1X, and devote some time in a separate patch for this specific case?

phil303: Do we know what APIs do this? Can we handle errors differently for them? This is going to bite us in the ass eventually. I really don't like abstracted out handling like this.

schwers: It's a bit messy, I did some grepping into r2 for the specific case the 1X comment mentions (a specific edge case when you do subreddit filtering via the url and the subreddits are invalid). That code is used in MinimalController and looks like it can be triggered on any endpoint that has a subreddit name in it. (Realistically its probably just the `r/<subreddit>/about.json` and `r/<subreddit>`). It looks like its an edge case specific to subreddit filtering that we don't support anyways on mweb at the moment, so I'd be in favor of just removing it all together for now and re-adding it in some other form later if it comes up again.

phil303:
:tropical_fish: Travis doesn't like it for some reason though

schwers:
Yeah, I really wish superagent made access to things like 'path' easy. What's frustrating is the threading feels like it should be unnecessary if we're already passing around `req`. I'd be happy to do the refactor this sprint, but I would rather do it in a separate patch after this. There's some error handling I'd like to rethink and there's going to be a lot of no-op changes for re-naming.

PR #141, opened 2016-08-12
Create simplified api request helper

phil303:
The motivation for this patch initially was to remove some of the abstraction from our api layer to gain me access to the http status codes. It then bloomed into a complete refactor of APIRequestUtils before I decided to reel the scope back in. What it is now is a single helper function that can do the work of `runForm`, `runQuery`, and `runJson` at a much lower abstraction level that deprecates those guys.

By design it does very little "handling" of the response for you but for consistency and simplicity does still hand you off an `ApiResponse` object. This object now has references to the initial request and its response so you can access any lower level things like the status codes should the need arise.

Ultimately, I'd like to give all of our api wrappers this treatment. The attempt to do that in this patch was just too much change in one go so I'll have to come back to it.

:eyeglasses: @schwers


apiBase/apiRequest.js
```diff
@@ -0,0 +1,36 @@
...
+  if (_method === 'post') {
+    request.send(body);
+  }
+
+  return new Promise((resolve, reject) => {
+    request.end((err, res) => {
+      if (!err) {
+        resolve(new APIResponse(res));
+      } else {
+        reject(err);
```
phil303: per convo with Ryan, we should reject `ResponseError` here to normalize superagent's error responses.

apiBase/apiRequest.js
```diff
@@ -0,0 +1,36 @@
+import superagent from 'superagent';
+
+import APIResponse from './APIResponse';
+
+
+export default (apiOptions, method, path, query={}, body={}, type=null) => {
```
phil303: consider making query, body, and type into one optional argument

phil303:
:haircut:

apiBase/apiRequest.js
```diff
@@ -0,0 +1,48 @@
...
+import APIResponse from './APIResponse';
+import ResponseError from './errors/ResponseError';
+
+
+/* A thin helper function around our apis.
+ *
+ * @param {Object} apiOptions - fields required to hit our api
+ * @param {String} method   - the http method upper or lowercased, e.g. 'GET'
+ * @param {String} path     - the endpoint path
+ * @param {String} options  - use to set query params, a body, or request type
```
nramadas: `* @param {Object} options`

phil303: fixing, thanks!

apiBase/apiRequest.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+export default (apiOptions, method, path, options={}) => {
+  const { query={}, body={}, type=null } = options;
+  const { origin, appName, env, token, defaultHeaders={} } = apiOptions;
+
+  const _method = method.toLowerCase();
+  const _headers = token ?
```
nramadas: for consistency, should we format this multiline ternary like we do the others?
```es6
const _header = token
  ? { ...defaultHeaders, Authorization: `Bearer ${token}` }
  : defaultHeaders;
```

phil303: good idea. updating.

apis/PostsEndpoint.js
```diff
@@ -60,41 +61,41 @@ const formatPostData = (data)=> {
...
 export default {
-  get(apiOptions, query) {
-    const path = getPath(query);
-    const apiQuery = formatQuery({...query}, 'get');
+  get(apiOptions, _query) {
+    const path = getPath(_query);
+    const query = formatQuery({..._query}, 'get');
 
-    return runQuery(apiOptions, 'get', path, apiQuery, query, handleGet);
+    return apiRequest(apiOptions, 'GET', path, { query }).then(handleGet);
```
nramadas: out of curiosity, why do you uppercase the methods?

apis/PostsEndpoint.js
```diff
@@ -60,41 +61,41 @@ const formatPostData = (data)=> {
...
+  return apiResponse;
 };
 
 export default {
-  get(apiOptions, query) {
-    const path = getPath(query);
-    const apiQuery = formatQuery({...query}, 'get');
+  get(apiOptions, _query) {
+    const path = getPath(_query);
+    const query = formatQuery({..._query}, 'get');
```
nramadas: the linter will probably catch this, but spaces to pad the `{}`

apis/PostsEndpoint.js
phil303: couple reasons - none i feel inherently strongly about.
1) Stands out. They could probably be constants but that makes the function slightly more unwieldy.
2) That's the way I'm used to seeing them, in the http spec, when I write stuff

nramadas:
:nail_care:

nramadas:
From a high level view, I think this is a great improvement. Just a few small nits, other than that it looks good.

apis/PostsEndpoint.js
phil303: ooo thank you

phil303:
thanks! updated :haircut:

nramadas:
:fish:

PR #145, opened 2016-08-17
Bad compilation

phil303:
Not sure what happened but when I tried to run the latest api-client on mweb, it was missing stuff. Re-running `node build` on api-client (rebased off latest) resulted in a diff so I must have missed a part of the compilation somehow. It's pretty hard to confirm since it's one big blob of uglified text though.

:eyeglasses: @nramadas


phil303:
Uzi managed to capture a good compilation in his latest patch. Closing.

PR #146, opened 2016-08-19
Post data fixes

uzi:
Takes care of a number of small issues... from fixing post sorting, to having data for properly displaying icons when posts are gilded, archived or locked.  There'll be some additional fixes on 2X, but this is a prerequisite.

👓 @phil303 @nramadas


phil303:
:fish:

PR #148, opened 2016-08-22
Query updates

uzi:
This takes care of both permalink support and passing raw_json=1 on GET requests, which was breaking a lot of things on the content side of the 2X experience.

👓  @schwers @phil303


apiBase/apiRequest.js
```diff
@@ -20,9 +20,13 @@ export default (apiOptions, method, path, options={}) => {
     ? { ...defaultHeaders, Authorization: `Bearer ${token}` }
     : defaultHeaders;
   const _path = path.startsWith('/') ? path : `/${path}`
-  const _query = { ...query, app: `${appName}-${env}` };
   const endpoint = origin + _path;
 
+  let _query = { ...query, app: `${appName}-${env}` };
+  if (_method === 'get') {
+    _query = { ...query, raw_json: 1 };
```
prashtx: It looks like this will remove the `app` param for GET requests. Is that intentional?

phil303: Just to put this in stone, as per our conversation, please do not merge this until `raw_json` is removed. It doesn't belong in `apiRequest` as this is agnostic to our endpoints. We're making the assumption that this should apply to every endpoint currently and going forward. Additionally, if we ever want to remove it, it becomes a very scary prospect. Finally, if we ever want to request something without `raw_json`, we won't be able to.

schwers:
:fish:

apiBase/apiRequest.js
uzi: @prashtx It wouldn't because of the `...` operator... but I've changed the implementation slightly based on discussion.  Thanks!

phil303:
:fish:

PR #149, opened 2016-08-24
Add `loadMoreIds` field to comment

phil303:
Open to suggestions on the name of this field. I also did some light cleaning up while digging through the code.


phil303:
:eyeglasses: @schwers

lib/commentTreeUtils.es6.js
```diff
@@ -51,30 +47,33 @@ function parseCommentData(data) {
...
+export function normalizeCommentReplies(comments, isTopLevel, visitComment) {
   return comments.map(comment => {
     if (comment.type === COMMENT_LOAD_MORE) { return; }
 
+    // assign some helpful keys and their defaults to the comment
+    Object.assign(comment, COMMENT_DEFAULTS);
+
     // Filter out if a comment is a "load more" type, set a property on the
     // parent comment, and then nuke the fake "reply"
-    if (comment.replies) {
```
phil303: i removed this check as I believe it to be redundant since we check to see all comments at this point should have a replies field.

models2/CommentModel.es6.js
```diff
@@ -27,6 +27,7 @@ export default class CommentModel extends RedditModel {
     replies: T.array,
     numReplies: T.number,
     loadMore: T.bool,
+    loadMoreIds: T.array,
```
schwers: 🔕 Another time I wish we had `T.arrayOf` ala React.PropTypes

phil303: put it on the [jira board](https://reddit.atlassian.net/secure/RapidBoard.jspa?rapidView=105&view=planning.nodetail&epics=visible&selectedEpic=MOBILEWEB-575)!

lib/commentTreeUtils.es6.js
schwers: Agreed, `parseCommentData` should be doing that

lib/commentTreeUtils.es6.js
```diff
@@ -6,7 +6,7 @@ import { COMMENT_LOAD_MORE } from '../models2/thingTypes';
 export function treeifyComments(comments=[]) {
   const commentDict = {};
   comments.forEach(c => {
-    commentDict[c.uuid] = c;
```
schwers: Is this another old bug? Guessing this was written assuming we got a model but it's really just `json` without the `uuid` set right?

apis/CommentsEndpoint.es6.js
```diff
@@ -53,16 +53,14 @@ const parseGetBody = (apiResponse, hasChildren) => {
...
-    } else {
-      comments = parseCommentList(body.json.data.things);
-    }
+
+    const { things } = body.json.data;
+    comments = treeifyComments(things.map(parseCommentData));
   }
 
-  normalizeCommentReplies(comments, (commentJSON, isTopLevel) => {
+  normalizeCommentReplies(comments, true, (commentJSON, isTopLevel) => {
```
schwers: 🔕  Minor style note, just a preference but in no way a blocker for this patch.  For recursive functions like this, I've always been a fan of having a helper that makes the first call with the correct starting value (in this case, that topLevel is `true`). Then you don't have to wonder what this unnamed `true` is, and you can't mess up calling it. We could still export the `_normalizeCommentReplies` function if we desire being able to call it manually.

models2/CommentModel.es6.js
schwers: Whoops, silly me, you can actually do `T.arrayOf` today! https://github.com/reddit/node-api-client/blob/3X/apiBase/Model.es6.js#L41

lib/commentTreeUtils.es6.js
phil303: At this point it's just the json and it's missing the uuid. It's a weird bug in that I can't imagine how it got into this state unless the field was there at some point.

schwers: I think treeify used to work on the models directly and got changed to json without proper testing. It will be a great day when we have tests in here

schwers:
:fish: lgtm, had some minor but non-blocker notes, I'll leave those to you

apis/CommentsEndpoint.es6.js
phil303: So one thing we can do is just to used named arguments, which is something I had but removed for some reason, probably line length issues. I can add it back in to get the same effect. Where possible, I generally like to avoid adding function layers. It makes reading the code just slightly more difficult.

models2/CommentModel.es6.js
phil303: Oh neat, I'll update.

PR #150, opened 2016-08-29
Pull in server loid into the Account model

nramadas:
The `api/me` call now returns `loid` and `loid_created` fields. This patch passes those along to the `Account` model.


nramadas:
:eyeglasses: @schwers

apiBase/APIRequestUtils.es6.js
```diff
@@ -43,7 +43,7 @@ const requestAuthHeader = (apiOptions) => {
 const requestHeaders = (apiOptions) => {
   const authHeaders = requestAuthHeader(apiOptions);
   return {
-    ...(apiOptions.defaultHeaders || {}),
+    ...(apiOptions.headers || {}),
```
phil303: nice thoroughness

phil303:
Not eyeglassed but this seems good to me. just in case you need it, :fish:

PR #151, opened 2016-09-08
Pass along loids generated by the api

nramadas:
Calls to /api/me return loids in the json response. This patch includes
those in the response body.

:eyeglasses: @schwers || @phil303

### DIFF
--- a/old-src/endpoints/v1.es6.js
+++ b/old-src/endpoints/v1.es6.js
@@ -1070,7 +1070,11 @@ class APIv1Endpoint {
           // user information.
           const data = {
             name: '',
-            ...(body.data || body)
+            ...(body.data || body),
+            loid: body.loid,
+            loid_created: typeof body.loid_created === 'number'
+              ? new Date(body.loid_created).toISOString()
+              : body.loid_created,
           };
           if (body) {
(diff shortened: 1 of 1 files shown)


old-src/endpoints/v1.es6.js
```diff
@@ -1070,6 +1070,10 @@ class APIv1Endpoint {
           // user information.
           const data = {
             name: '',
+            loid: body.loid,
+            loid_created: typeof body.loid_created === 'number'
```
nramadas: right now, we get back a time in ms. this seems incompatible with our current `loidcreated` cookies, and throws errors when we try to use them. i'll make a ticket to get this changed, but for now, this is a safety check.

old-src/endpoints/v1.es6.js
```diff
@@ -1070,6 +1070,10 @@ class APIv1Endpoint {
...
+              ? new Date(body.loid_created).toISOString()
+              : body.loid_created,
             ...(body.data || body)
```
schwers: from our conversation: this should be moved up, just to be safe

schwers:
:fish:

PR #153, opened 2016-09-13
Don't accept a thumbnail value of "image"

uzi:
Without this, we try to load https://m.reddit.com/image

👓  @schwers


schwers:
:fish:

PR #155, opened 2016-09-15
remove response parsing timing from log

phil303:
I don't think anyone is actively using it and it's just white noise/cluttering up the logs.

:eyeglasses: @nramadas @schwers


schwers:
Yeah, this was helpful in dev a while ago, but it's time to go. :fish:

PR #156, opened 2016-09-16
ValidationError was not being imported

phil303:
ValidationError was getting called without being imported so import it.

:eyeglasses: @nramadas || @schwers


nramadas:
:fish:

schwers:
:fish:

PR #157, opened 2016-09-16
Make checking in and publishing code simpler/more correct

phil303:
Problem:
Currently we check-in our transpiled code. Other than this being a
general nuisance, it's actually a little dangerous as well. Because it's
essentially one big blob of code, it's hard to read and parse. We've
already encountered situations when someone's checked in code bashed
someone else's changes.

Fix:
It's fairly simple to fix this by simply checking in only the source
code and publishing only the build code.

Details:
I've moved all our source code into /src, that way it's easier to target
for things like npmignore and our linter. This change is actually
straightforward since all the paths are relative and therefore I only
needed to update webpack's config to point to /src/index.es6.js.

I then tell npmignore to ignore almost everything but our build file
(and some other small files that I wasn't too sure about - repl and
run.js).

Finally I updated package.json to have a prepublish hook to ensure that
our source code is transpiled before it's pushed to npm.

Other changes:

- Removed a bunch of gitignore things that looked irrelevant to this
  library
- Removed .editorconfig and mancyLoader, which also looks irrelevant.
- Updated package.json with more up to date information and updated
  build to 0.10.0 (tested this out as well).

There's some other funky stuff going on in this repo, such as js files
being executable and the .es6.js extensions but in the interest of
smaller changes (and this is already a pretty big one), I'm not going to
touch those on this pass.


phil303:
:eyeglasses: @schwers || @nramadas || @griffinmichl

schwers:
🔕  Part of me thinks we should keep the source and lint files in the repo, I find myself changing that at times for quick edits. I also want to setup building this source directly into our consumers instead of using the built version (for tree shaking / sourcemaps). We can easily change that once the later is setup.

🐟  Overall I think this is a big improvement

phil303:
@schwers I'm actually not sure what the "correct" thing to do there is. Just a quick audit of our libraries seems to indicate it's normal to publish the source. I can't really think of a reason not to except I thought someone at some point told me that it wasn't the correct thing to do and payload size.

schwers:
@phil303 I think overall keeping the source is "the right thing to do" :tm:. The first thing that comes to mind to me is building the source yourself for advanced users. I think it makes diff of the dependencies repo a little more parseable as well, but I care less about that as we can usually track module version numbers to commits pretty easily

phil303:
For posterity, I discussed this with @schwers. Apparently leaving the source code makes folks using rollup's lives easier so for now, I'm going to add it back in.

phil303:
:haircut: @schwers

schwers:
:fish:

PR #158, opened 2016-09-20
Fix bad export statements

phil303:
Problem:
This seems to have been unearthed when webpack 2 got updated. Things
that were not default exported were being imported as such.

Fix:
For ApiResponse, make it a non-default import and adjust all imports
accordingly. For the HiddenPostsAndComments collection, make it the
default. This clears up the warnings in the build step.

:eyeglasses: @nramadas || @uzi

### DIFF
--- a/src/apiBase/APIResponse.es6.js
+++ b/src/apiBase/APIResponse.es6.js
@@ -85,7 +85,7 @@ export class APIResponseBase {
   appendResponse() { throw new Error('Not implemented in base class'); }
 }
 
-export default class APIResponse extends APIResponseBase {
+export class APIResponse extends APIResponseBase {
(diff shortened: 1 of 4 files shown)


src/apiBase/APIResponse.es6.js
```diff
@@ -85,7 +85,7 @@ export class APIResponseBase {
   appendResponse() { throw new Error('Not implemented in base class'); }
 }
 
-export default class APIResponse extends APIResponseBase {
+export class APIResponse extends APIResponseBase {
```
phil303: I figured APIResponse was no more important than the other classes in this file so I made it a non-default. I can flip this if we disagree.

uzi:
:fish:

PR #160, opened 2016-09-28
Fix babel error with super and writing readonly props

schwers:
There are two big fixes in this patch, the first is a
simple issue with readonly props. FakeError has readonly
`message` and `stack` properties that were overwitten with
calls to `Object.assign`. The fix is an alternative helper
that won't overrwrite those props.

The second fix is for a bug in babel. The ResponseError
constructor tries to bail out and return a different
FakeError subclass to differentiate Disconnection Errors.
Babel was inserting its guard function to ensure super was
called before the constructor exits (which we don't need
because we bail out explicitly).

This fixes an error in mweb 2X when we try to create a new `ResponseError` in `src/app/models/Session.js:fetchLogin`. You can add this to the `.end` call there to see it in action:

```js
  console.log('forcing response error', err, 'err has code?', err.code);
  const resultError = new ResponseError({
    message: 'connection refused',
    status: 500,
    syscall: true,
    code: 'ECONNREFUSED',
  }, '/loginproxy');

  console.log('result error is?', resultError);
  return reject(resultError);
```

Here's a babel repl 'gist' so you can see the babel compilation bug that this patch fixes in action [look at compiled code line 24](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=class%20Foo%20%7B%0A%20%20constructor(thing)%20%7B%0A%20%20%20%20this.thing%20%3D%20thing%3B%0A%20%20%7D%0A%7D%0A%0Aclass%20Bar%20extends%20Foo%20%7B%0A%20%20constructor%20(thing%2C%20otherThing)%20%7B%0A%20%20%20%20if%20(otherThing)%20%7B%0A%20%20%20%20%20%20return%20new%20Baz()%3B%0A%20%20%20%20%7D%0A%20%20%20%20%0A%20%20%20%20super(thing)%3B%0A%20%20%7D%0A%7D%0A%0Aclass%20Baz%20extends%20Foo%20%7B%20%7D%0A%0A)

👓  @nramadas  @uzi

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "superagent": "1.x"
   },
   "devDependencies": {
-    "@r/build": "~0.10.0",
+    "@r/build": "~0.10.1",
     "babel-core": "^6.8.0",
     "babel-eslint": "^6.0.4",
     "babel-plugin-syntax-trailing-function-commas": "^6.8.0",
--- a/src/apiBase/errors/FakeError.es6.js
+++ b/src/apiBase/errors/FakeError.es6.js
@@ -1,3 +1,5 @@
+import omit from 'lodash/omit';
+
 export default class FakeError {
   constructor (message) {
     Object.defineProperty(this, 'message', { value: message });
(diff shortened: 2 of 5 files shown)


uzi:
👍  from me and I'll give @nramadas a chance to have a peek

phil303:
:fish:

nramadas:
:fish:

PR #163, opened 2016-10-05
Message GET updates - replies and threads

uzi:
Store all of the replies when we get all of the
threads and add support for fetching just one
message thread.

👓  @schwers @phil303

### DIFF
--- a/src/apis/MessagesEndpoint.es6.js
+++ b/src/apis/MessagesEndpoint.es6.js
@@ -11,7 +11,18 @@ const TYPE_MAP = {
 
 const parseGetBody = (res, apiResponse, apiRequest) => {
   const { body } = res;
-  body.data.children.forEach(c => apiResponse.addResult(TYPE_MAP[c.kind].fromJSON(c.data)));
+  body.data.children.forEach(c => {
+    const replies = [];
+    if (c.data.replies) {
+      c.data.replies.data.children.forEach(r => {
+        apiResponse.addModel(TYPE_MAP[r.kind].fromJSON(r.data));
+        replies.push(`${r.kind}_${r.data.id}`);
+      });
+    }
+
+    c.data.replies = replies;
+    apiResponse.addResult(TYPE_MAP[c.kind].fromJSON(c.data));
+  });
 };
 
 const parsePostBody = (res, apiResponse, apiRequest) => {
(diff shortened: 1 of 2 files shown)


uzi:
👓  @nramadas

src/apis/MessagesEndpoint.es6.js
```diff
@@ -20,9 +29,10 @@ const parsePostBody = (res, apiResponse, apiRequest) => {
 };
 
 const getPath = (query={}) => {
-  const { subreddit, type } = query;
+  const { subreddit, type, thread } = query;
   const sub = subreddit ? `r/${ subreddit }/` : '';
-  return `${ sub }message/${ type }.json`;
+  const id = thread ? `/${ thread }` : '';
+  return `${ sub }message/${ type }${ id }.json`;
```
phil303: I think you're probably just being consistent within this file but I don't think we normally put the spaces between the variable and the curlies while doing string interpolation. Can you remove them here and on line 33?

phil303:
One nitpick comment that I don't really feel that strongly about. Code looks ok and I'll trust you that it works :fish:!

src/apis/MessagesEndpoint.es6.js
```diff
@@ -11,7 +11,16 @@ const TYPE_MAP = {
 
 const parseGetBody = (res, apiResponse, apiRequest) => {
   const { body } = res;
-  body.data.children.forEach(c => apiResponse.addResult(TYPE_MAP[c.kind].fromJSON(c.data)));
+  body.data.children.forEach(c => {
+    const replies = [];
+    c.data.replies && c.data.replies.data.children.forEach(r => {
```
schwers: 🔕 Its interesting to see our usage of `&&` for minor control flow in JSX evolve, now we're using it in api code

schwers:
:fish:

src/apis/MessagesEndpoint.es6.js
phil303: yea actually. that's a good point that I didn't really notice before. Maybe we shouldn't do that?

uzi: Oh, hah!  I had it both ways and went with this but I can put it in an if statement if it makes you more comfortable.

PR #164, opened 2016-10-05
Expose apiRequest

phil303:
I forgot to expose this in the original patch and I'm using it in mweb.

:eyeglasses: @schwers || @nramadas


phil303:
:eyeglasses: @schwers || @nramadas

PR #165, opened 2016-10-05
Replace captcha with recaptcha for mweb

jackniu1:
👓 @uzi @KeyserSosa 

Have to wait for below PR deployed first:

https://github.com/reddit/reddit-public/pull/5828

### DIFF
--- a/src/apis/PostsEndpoint.js
+++ b/src/apis/PostsEndpoint.js
@@ -52,7 +52,7 @@ const formatPostData = (data)=> {
     sendreplies: data.sendreplies,
     sr: data.sr,
     iden: data.iden,
-    captcha: data.captcha,
+    'g-recaptcha-response': data.gRecaptchaResponse,
     resubmit: data.resubmit,
   };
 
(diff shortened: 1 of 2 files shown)


src/apis/PostsEndpoint.js
```diff
@@ -52,7 +52,7 @@ const formatPostData = (data)=> {
     sendreplies: data.sendreplies,
     sr: data.sr,
     iden: data.iden,
-    captcha: data.captcha,
+    'g-recaptcha-response': data.gRecaptchaResponse,
```
uzi: What is `iden` and do we still need it after switching to recaptcha?

jackniu1: Yep, it was used to identify whether the recaptcha need show or not. Will send another PR of reddit-mobile soon.

KeyserSosa:
:fish: assuming I shouldn't be worried about the unit test!

uzi:
:fish: as well ... but of course have to wait for that other patch

PR #166, opened 2016-10-06
Add support for EditUserText endpoint

schwers:
This api enables editing posts / comments. I needed to expose the raw markdown as part of this so that the 2X client can actually edit that

I'll squash all the `[wip]` commits into one commit when this is done with review and give it a proper message.

👓  @phil303  || @uzi  || @nramadas


phil303:
one comment but seems good overall :nail_care:

src/apis/EditUserTextEndpoint.js
```diff
@@ -0,0 +1,45 @@
...
+        text,
+        thing_id: thingId,
+      },
+    };
+
+    return apiRequest(apiOptions, 'POST', ENDPOINT, options)
+      .then(apiResponse => {
+        const { body: { json }} = apiResponse.response;
+        if (json.errors.length) {
+          throw json.errors[0];
```
schwers: I'm not sure. Private throws the json errors array. I could use the `ValidationError` class, but I don't have a strong preference, do you have a preference?

uzi:
Same thoughts as @phil303

src/apis/EditUserTextEndpoint.js
phil303: Hey ryan, I'm not sure what you mean by "private". I don't have any strong preferences here but I thought our general practice was to use these custom error instances so it makes type checking more straightforward/explicit for [callers](https://github.com/reddit/reddit-mobile/pull/804/files#diff-9f7e2144dca9d72611cd8d6c7fde04a1R106) of these apis. It does seems like ValidationError would be the right thing though.

schwers: Ah by private I mean `node-private`. What throws me a little about `ValidationError` is the name sounds like a client-side validation issue to me. I think overall its the best approach though, I'll update.

schwers:
:haircut_man:

phil303:
:fish:

PR #172, opened 2016-10-28
Fix comment and post downvotes

phil303:
Bug:
We get `true`, `false`, and `null` back from the api to represent upvote,
downvote, and no vote respectively. Currently the code is trying to
coerce `false` to -1. This doesn't work since it actually gets coerced to
0. So downvotes from the api were not being represented in the ui.

Fix:
Use a switch statement to normalize the api values to 1, 0, and -1. Also
name the validator `likes` since this is a super specific validator.

Additional changes:
There was an additional potential bug from a recent patch. Since the
api can return falsey values, we don't want to check for falseyness as
a way to determine if we need the fallback value. Although this happens
to work out currently, its ambiguity is probably dangerous.

The fix for this is to check explicitly for `undefined` since this
implies a missing key. Any other value value means it was explicitly set.

:eyeglasses: @schwers

### DIFF
--- a/src/apiBase/Model.es6.js
+++ b/src/apiBase/Model.es6.js
@@ -40,9 +40,15 @@ export default class Model {
     array: val => Array.isArray(val) ? val : [],
     arrayOf: (type=Model.Types.nop) => val => Model.Types.array(val).map(type),
     bool: val => Boolean(val),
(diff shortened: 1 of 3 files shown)


schwers:
:fish:

PR #173, opened 2016-11-01
Model fixes

uzi:
This fixes a bit of data that api-client sends back. Note that there should be a few more fixes for the mail stuff before this goes in:

1) We'll want to POST to [/api/read_message](https://www.reddit.com/dev/api/oauth#POST_api_read_message) ids of messages as we read them.  This can be done in reddit-mobile, but it might be better to do here in node-api-client in MessagesEndpoint.es6.js ... probably in the get() method when there's a specific thread we're fetching.
2) We'll want to update reddit-mobile's messaging views to highlight (or embolden) messages that are new.
3) We'll need some `badge` css styling for reddit-mobile in the `SettingsOverlayMenu` where we show `inboxCount`.

I consider this a WIP until those things are done as well.

### DIFF
--- a/src/models2/Account.es6.js
+++ b/src/models2/Account.es6.js
@@ -33,7 +33,7 @@ export default class Subreddit extends RedditModel {
   }
 
   static API_ALIASES = {
-    comment_karm: 'commentKarma',
+    comment_karma: 'commentKarma',
     created_utc: 'createdUTC',
     gold_creddits: 'goldCreddits',
     gold_expiration: 'goldExpiration',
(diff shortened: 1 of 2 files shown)


src/models2/Account.es6.js
```diff
@@ -33,7 +33,7 @@ export default class Subreddit extends RedditModel {
   }
 
   static API_ALIASES = {
-    comment_karm: 'commentKarma',
+    comment_karma: 'commentKarma',
```
phil303: good find. curious what this was breaking

phil303:
@uzi so is this just a prep patch?

uzi:
Yeah... just wanted to get these things captured in case someone else has a chance to take a stab at any of the three things I mentioned... or at leas to have the discussion with folks.

PR #176, opened 2016-11-17
Continue thread

curioussavage:
I added this prop because Phil requested it in https://github.com/reddit/reddit-mobile/pull/721

Also the NPM was exiting before initial install saying the superagent-retry repo was not there.

👓  @schwers @uzi


schwers:
:fish:

PR #177, opened 2016-11-21
Add queryParams to apiOptions

schwers:
APIOptions has a configurable headers field, clients use this
to add headers to api requests. This patch adds similiar functionality
but for queryParams.

👓  @curioussavage @nramadas


curioussavage:
looks good, tested locally too. 🐟

PR #180, opened 2016-12-12
Add disableComments attribute to post model

jamesmaa:
👓 @schwers 

We're seeing some bugs where a PostObject has `num_comments` set but the post has comments disabled (e.g. some ad units). We need to surface this attribute to mweb.


schwers:
d'oh, good catch @jamesmaa :fish:

PR #185, opened 2017-01-12
Bump version to 3.32.1

birakattack:
Bumping version to 3.32.1 to fix issue where the wrong commit was tagged.

:eyeglasses: @schwers


schwers:
:fish:

PR #188, opened 2017-01-27
Add distingiush endpoint to modtools

scarow:
👓 @birakattack @madbook

### DIFF
--- a/src/apis/modTools.es6.js
+++ b/src/apis/modTools.es6.js
@@ -1,19 +1,34 @@
 import apiRequest from '../apiBase/apiRequest';
 
-const remove = (apiOptions, fullname, spam) => {
-    // Remove a link, comment, or modmail message.
-    const body = {
-    	id: fullname,
-    	spam: spam
-    };
+const DISTINGUISH_TYPES = {
+  NONE: '',
+  MODERATOR: 'moderator',
+  ADMIN: 'admin',
+};
 
-    return apiRequest(apiOptions, 'POST', 'api/remove', { body, type: 'form' });
+const remove = (apiOptions, fullname, spam) => {
+  // Remove a link, comment, or modmail message.
+  const body = {
+    id: fullname,
+    spam: spam
+  };
+
+  return apiRequest(apiOptions, 'POST', 'api/remove', { body, type: 'form' });
 }
 
 const approve = (apiOptions, fullname) => {
-    // Approve a link or comment
-    const body = { id: fullname };
-    return apiRequest(apiOptions, 'POST', 'api/approve', { body, type: 'form' });
+  // Approve a link or comment
+  const body = { id: fullname };
+  return apiRequest(apiOptions, 'POST', 'api/approve', { body, type: 'form' });
 }
 
-export default { remove, approve }
+const distinguish = (apiOptions, fullname, distinguishType) => {
+  // Distinguish a link or comment
+  const body = {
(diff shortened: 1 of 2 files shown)


madbook:
[relevant api docs for reference](https://www.reddit.com/dev/api/#POST_api_distinguish)

src/apis/modTools.es6.js
```diff
@@ -1,19 +1,34 @@
 import apiRequest from '../apiBase/apiRequest';
 
+const DISTINGUISH_TYPE = {
+  'moderator': 'yes',
+  '': 'no',
+  'admin': 'admin',
+};
+
```
scarow: This feels a little weird. Would rather do something like:
```
{
  MODERATOR: 'yes',
  NONE: 'no'
}
```

But in the client app we are passing around `'moderator'` and `''` strings (and passing those strings to node-api-client). Maybe there is some special javascript syntax I can use?

scarow:
:haircut_man:

src/apis/modTools.es6.js
```diff
@@ -1,19 +1,34 @@
...
+  const body = {
+    how: DISTINGUISH_TYPE[distinguishType],
+    id: fullname,
+  };
+  return apiRequest(apiOptions, 'POST', 'api/distinguish', { body, type: 'form' });
 }
 
-export default { remove, approve }
\ No newline at end of file
+export default { remove, approve, distinguish, DISTINGUISH_TYPE }
```
birakattack: Why are we exporting `DISTINGUISH_TYPE`?

src/apis/modTools.es6.js
```diff
@@ -1,19 +1,34 @@
...
-    return apiRequest(apiOptions, 'POST', 'api/approve', { body, type: 'form' });
+  // Approve a link or comment
+  const body = { id: fullname };
+  return apiRequest(apiOptions, 'POST', 'api/approve', { body, type: 'form' });
+}
+
+const distinguish = (apiOptions, fullname, distinguishType) => {
+  // Distinguish a link or comment
+  const body = {
+    how: DISTINGUISH_TYPE[distinguishType],
```
birakattack: Should we check if the passed `distinguishType` is a valid value? Maybe we can just default to `null` and just let the api return an error?

src/apis/modTools.es6.js
madbook: I think ideally we'd export `DISTINGUISH_TYPE` here as it's intended to be used in the client app, e.g.

```javascript
/**
  * @enum
  */
export DISTINGUISH_TYPE = {
  NONE: '',
  MODERATOR: 'moderator',
  ADMIN: 'admin',
};
```

and import that rather than redefining it in the client app.  Then, handle the mapping of `DISTINGUISH_TYPE.MODERATOR → 'yes'` and `DISTINGUISH_TYPE.NONE → 'no'` for the API call within the `distinguish` function.

src/apis/modTools.es6.js
scarow: Oh -- whoops! remnants of a different idea :P

scarow:
I actually think this is a remnant of what we were going to implement at first @madbook. Now, I am only passing around `moderator` and `''` in the client app. When it hits node-api-client, that's when I use the hash values.

I can delete the export.

scarow:
:haircut_man:

PR #189, opened 2017-01-27
Fix repl

madbook:
It looks like `apiClient.js` was replaced with `build.js` [here](https://github.com/reddit/node-api-client/commit/f9ff73a14b8858b91d7d1f4cc75a5402a53c195a), but this spot in `repl` wasn't updated.

👓 @schwers


schwers:
:fish: d'oh, thanks for catching this!

madbook:
12ebc4a76bb00cad715a272e2506dceac57c7175 included in release v3.34.0

PR #191, opened 2017-01-30
Nsfw, spoiler, and lock post actions

birakattack:
Integrated nsfw, spoiler and lock endpoints into the modTools api.

:eyeglasses: @madbook @scarow

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@r/api-client",
-  "version": "3.34.0",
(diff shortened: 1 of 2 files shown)


birakattack:
The build shows as a failure if you pushed a commit that had an issue then subsequently fixed it. If you actually click on the failure you'll see the latest build completed successfully - I'll let infra know about it as its pretty annoying.

PR #192, opened 2017-01-30
Modtools: Add distinguishType map

scarow:
👓 @birakattack @madbook 

Update: moved distinguishType mappings into node-api-client. Previously they were in the client, but this seems cleaner/better.

### DIFF
--- a/src/apis/modTools.es6.js
+++ b/src/apis/modTools.es6.js
@@ -24,8 +24,14 @@ const approve = (apiOptions, fullname) => {
 
 const distinguish = (apiOptions, fullname, distinguishType) => {
   // Distinguish a link or comment
+
+  const distinguishTypeMap = {
+    'moderator': 'yes',
+    '': 'no',
+  };
+
   const body = {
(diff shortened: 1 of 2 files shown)


src/apis/modTools.es6.js
```diff
@@ -24,8 +24,14 @@ const approve = (apiOptions, fullname) => {
...
   // Distinguish a link or comment
+
+  const distinguishTypeMap = {
+    'moderator': 'yes',
+    '': 'no',
+  };
+
   const body = {
-    how: distinguishType,
+    how: distinguishTypeMap[distinguishType],
```
birakattack: Should we add a check here to ensure that the passed `distinguishType` is valid?

scarow: Oh! yes :)

scarow: Eh -- the toaster in the UI takes care of errors here.

scarow:
:haircut_man:

PR #194, opened 2017-01-31
Add setStickyComment to api/modTools

madbook:
Enabled setting sticky comments.

The API for sticky comments is a little weird – it's actually an optional boolean flag sent to the distinguish endpoint used to distinguish comments and posts.   It only works for comments though (sticky posts have a separate API endpoint to handle some of their additional complexity).

I initially had trouble getting the distinguish API to work with the `sticky` flag.  I discovered that the `how` argument is actually meant to be [passed as part of the _path_](https://github.com/reddit/reddit-public/blob/master/r2/r2/config/routing.py#L427), rather than in the post body.  Moving it allows the call to work for both use cases.  I have _no idea_ why it works w/o the sticky flag (i.e. for normal distinguish/undistinguish calls).

👓 @scarow @birakattack

### DIFF
--- a/src/apis/modTools.es6.js
+++ b/src/apis/modTools.es6.js
@@ -1,5 +1,11 @@
 import apiRequest from '../apiBase/apiRequest';
 
+/**
+ * Valid distinguish types.
+ * Note that the API endpoint used to distinguish posts and comments accepts
+ * 'yes' instead of 'moderator' and 'no' instead of ''.  See #distinguish
+ * @enum
+ */
 const DISTINGUISH_TYPES = {
   NONE: '',
   MODERATOR: 'moderator',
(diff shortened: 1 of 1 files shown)


src/apis/modTools.es6.js
```diff
@@ -22,19 +28,46 @@ const approve = (apiOptions, fullname) => {
...
-
+/**
+ * Distinguish a link or comment
+ * @function
+ * @param {Object} apiOptions
+ * @param {string} fullname The fullname of the target comment
+ * @param {DISTINGUISH_TYPES} distinguishType What type of distinguish is being set
+ * @param {?bool} [_sticky] For internal use by #setStickyComment
+ */
+const distinguish = (apiOptions, fullname, distinguishType, _sticky=null) => {
```
scarow: Thanks for adding this stuff!

src/apis/modTools.es6.js
```diff
@@ -22,19 +28,46 @@ const approve = (apiOptions, fullname) => {
...
+    [DISTINGUISH_TYPES.MODERATOR]: 'yes',
+    [DISTINGUISH_TYPES.NONE]: 'no',
   };
 
   const body = {
-    how: distinguishTypeMap[distinguishType],
     id: fullname,
   };
-  return apiRequest(apiOptions, 'POST', 'api/distinguish', { body, type: 'form' });
+  const how = distinguishTypeMap[distinguishType] || distinguishType;
```
scarow: Why the `||`?

scarow:
💅 looks good just a question!

src/apis/modTools.es6.js
madbook: since `admin` is defined in DISTINGUISH_TYPES, to support it we either need to define it in `distinguishTypeMap`, or just add a fallback to use the value directly (what the `|| distinguishType` is doing)

madbook:
:haircut_man: on the question

src/apis/modTools.es6.js
scarow: cool! sounds good 2 meee

scarow:
🐟

madbook:
926bd811a0fa2732a1b88645f41ee22a8f0fe6ae included in release v3.34.0

PR #195, opened 2017-02-01
Add rules support

madbook:
API support for [CE-462](https://reddit.atlassian.net/browse/CE-462).  The ticket only really requires the `get` method support, but it wasn't much extra effort to add `post`, `put`, and `del` as well. This also cleans out some old rules code that wasn't actually getting used anywhere.

👓 @birakattack @scarow


src/apis/SubredditRulesEndpoint.es6.js
```diff
@@ -0,0 +1,125 @@
...
+    const body = {
+      api_type: 'json',
+      raw_json: 1,
+      r: subredditName,
+      description: data.description,
+      kind: data.kind,
+      short_name: data.shortName,
+    };
+
+    return apiRequest(apiOptions, { body, type: 'form' });
```
birakattack: I believe you are not passing the `path`, and the request method here (`POST`).

src/apis/SubredditRulesEndpoint.es6.js
```diff
@@ -0,0 +1,125 @@
...
+      api_type: 'json',
+      raw_json: 1,
+      r: subredditName,
+      old_short_name: shortName,
+      description: data.description,
+      kind: data.kind,
+      short_name: data.shortName,
+    };
+
+    return apiRequest(apiOptions, { body, type: 'form' });
```
birakattack: Need to pass the `path` and request `method` as well here.

src/apis/SubredditRulesEndpoint.es6.js
```diff
@@ -0,0 +1,125 @@
...
+  async del(apiOptions, subredditName, shortName) {
+    const path = REMOVE_RULE_PATH;
+    const body = {
+      api_type: 'json',
+      raw_json: 1,
+      r: subredditName,
+      short_name: shortName,
+    };
+
+    return apiRequest(apiOptions, { body, type: 'form' });
```
birakattack: Need to pass the `path` and request `method` as well here.

src/apis/SubredditRulesEndpoint.es6.js
```diff
@@ -0,0 +1,125 @@
...
+   * @param {Object} apiOptions
+   * @param {string} subredditName The name of a subreddit
+   * @param {Object} data
+   * @param {string} data.description Markdown formatted description of the rule
+   * @param {RULE_TARGET} data.kind The types of things the rule applies to
+   * @param {string} data.shortName A short, plaintext title for the rule
+   */
+  async post(apiOptions, subredditName, data) {
+    const path = ADD_RULE_PATH;
+    const body = {
```
birakattack: Should we be doing some input validation in here?

src/models2/SubredditRule.es6.js
```diff
@@ -0,0 +1,34 @@
...
+export default class SubredditRule extends RedditModel {
+  static type = SUBREDDIT_RULE;
+
+  static PROPERTIES = {
+    createdUTC: T.number,
+    description: T.string,
+    descriptionHTML: T.string,
+    kind: T.string,
+    priority: T.number,
+    shortName: T.string,
```
birakattack: Do any of these properties need to be set to `required`?

src/models2/SubredditRule.es6.js
```diff
@@ -0,0 +1,34 @@
...
+  static API_ALIASES = {
+    short_name: 'shortName',
+    created_utc: 'createdUTC',
+    description_html: 'descriptionHTML',
+  };
+
+  makeUUID(data) {
+    // The actual rules model in r2 doesn't have a proper unique key, but
+    // the `created_utc` timestamp should work since it shouldn't change.
+    return `${data.subreddit}/${data.createdUTC}`;
```
birakattack: Should this be `data.subredditName` instead of `data.subreddit`?

src/models2/thingTypes.es6.js
```diff
@@ -24,6 +24,9 @@ export const PROMOCAMPAIGN_TYPE = 't8';
 export const WIKI = 'wiki';
 export const WIKI_TYPE = 'wiki';
 
+export const SUBREDDIT_RULE = 'subreddit_rule';
+export const SUBREDDIT_RULE_TYPE = 'subreddit_rule';
```
birakattack: Just to confirm subreddit rules are not considered a `thing`?

src/apis/SubredditRulesEndpoint.es6.js
madbook: oh jeeze, what was I doing over here

src/models2/SubredditRule.es6.js
madbook: oh good catch – i changed that property name last minute and missed this!

src/models2/SubredditRule.es6.js
madbook: Is that a thing in this context? (these aren't react proptype validators, they are [these things](https://github.com/reddit/node-api-client/blob/3X/src/apiBase/Model.es6.js#L37-L61)

src/models2/thingTypes.es6.js
madbook: yeah, one of the few things that aren't `things`

madbook:
:haircut_man: :face-palm: I can't believe I left those params out, lol

src/models2/SubredditRule.es6.js
birakattack: whoops - sorry the context switching broke my brain 👍

src/apis/SubredditRulesEndpoint.es6.js
madbook: I'm not sure how we're doing it generally at the moment – we could duplicate the validation logic locally here, or rely on r2 and let the api wrapper be dumb.  The latter is generally my preferred approach here since it means one less place to update, but if we are typically duplicating the validation for APIs locally then maybe it makes sense for consistency.

In either case, nothing will currently be using these (the subreddit rules stulff I'm working on is only using the `get` endpoint), so kicking it down the road is also an option :P

madbook:
also :face-palm: because i 👓 'd the wrong sam initially... @scarow

scarow:
haha -- i get pinged on `sam` anyways so i've seen it!

src/models2/SubredditRule.es6.js
madbook: I know... I really wish these didn't look like react proptype validators... :P

madbook:
Added a couple commits (75422ac  and 0260b96 ) to deal with small issues i ran into on the mweb side, nothing major.

src/apis/SubredditRulesEndpoint.es6.js
birakattack: That works for me - was more curious to get your opinion. :+1:

madbook:
:haircut_man: just for inclusion of d926c18

madbook:
0568a6232bba5879d29ec2a0656008bebc9281c3...e01bd1d24ac899349f28a6158fea38cc77d19d36

PR #197, opened 2017-02-04
Comments: make comments a flat list

curioussavage:
To enable this I had to make new models for load more and continue thread fake comments.
These are now left in the flat list instead of using flags/data on the parent
to know when to show them. I tried to optimize the methods for processing
comment results. Worst case would be for loadMore responses we have to iterate through
the list twice to build the tree (to get depth more easily) then we iterate and walk
the comment trees once to build the instantiate the models, build the
flat list of comment stubs and create the dictionary of <comment id> =>
<comment data>.

### DIFF
--- a/src/apiBase/Record.es6.js
+++ b/src/apiBase/Record.es6.js
@@ -5,3 +5,10 @@ export default class Record {
     this.paginationId = paginationId;
   }
 }
+
+export class DepthRecord extends Record {
+  constructor(type, uuid, paginationId=uuid, depth) {
+    super(type, uuid, paginationId);
+    this.depth = depth;
+  }
+}
--- a/src/models2/CommentModel.es6.js
+++ b/src/models2/CommentModel.es6.js
@@ -1,5 +1,5 @@
 import RedditModel from './RedditModel';
-import Record from '../apiBase/Record';
+import { DepthRecord } from '../apiBase/Record';
 import { COMMENT, COMMENT_LOAD_MORE } from './thingTypes';
 
 import votable from './mixins/votable';
@@ -90,9 +90,10 @@ export default class CommentModel extends RedditModel {
   };
 
   toRecord() {
-    const record = new Record(this.type, this.uuid, this.paginationId);
-    record.depth = this.depth;
-    return record;
+    return new DepthRecord(this.type,
+                           this.uuid,
+                           this.paginationId,
+                           this.depth);
   }
 }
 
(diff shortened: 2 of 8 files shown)


src/models2/ContinueModel.es6.js
```diff
@@ -0,0 +1,30 @@
...
+    id: T.string,
+    depth: T.number,
+  }
+
+  static API_ALIASES = {
+    parent_id: 'parentId',
+  }
+
+  makeUUID() {
+    return (Math.random() * 16).toFixed();
```
schwers: Could this be `${this.parentId}-${COMMENT_CONTINUE_THREAD}`? Would be nice to make this deterministic, and avoid the (low, but still possible) chance of collision.

src/models2/LoadMoreModel.es6.js
```diff
@@ -0,0 +1,27 @@
...
+    parentId: T.string,
+    children: T.arrayOf(T.string),
+    count: T.number,
+    depth: T.number,
+  }
+
+  static API_ALIASES = {
+    parent_id: 'parentId',
+  }
+
```
schwers: 🔕  Should this have a `makeUUID` function?

src/models2/CommentModel.es6.js
```diff
@@ -89,30 +87,12 @@ export default class CommentModel extends RedditModel {
...
+  };
 
   toRecord() {
-    if (this.uuid === this.name) {
-      return super.toRecord();
-    }
-
-    // otherwise its a load more stub for super nested comments
-    return new Record(COMMENT_LOAD_MORE, this.parentId);
+    const record = new Record(this.type, this.uuid, this.paginationId);
```
schwers: Let's make a subclass of `Record` that has `depth` set in the constructor

src/models2/LoadMoreModel.es6.js
curioussavage: load more has an id/name from api which happens to actually be the id its first child. yep..

curioussavage:
:haircut_man:

schwers:
👓  @schwers

src/apiBase/Record.es6.js
```diff
@@ -5,3 +5,10 @@ export default class Record {
     this.paginationId = paginationId;
   }
 }
+
+export class DepthRecord extends Record {
+  constructor(type, uuid, paginationId=uuid, depth) {
+    super(type, uuid, paginationId);
+    this.depth = depth;
+  }
+}
```
schwers: :nit: needs an eof newline

